Issue #51, opened 2017-06-08
gulp-inline-ng2-template: Error with @angular/animations


ChrisProlls:
Because of lack of publishing lib with Angular CLI, I started to use your plugin which packages our Angular app into NPM package  that can be published. Our whole Angular application is both library and application that can be built to be hosted. My goal here is to package the application so that others applications can use our reusable components. There is the `ng-package.json` : 
```JSON
{
  "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
  "src": "",
  "dest": "dist",
  "workingDirectory": ".ng_build",
  "lib": {
    "entryFile": "src/public_api.ts"
  }
}
```
The `public_api.ts` only export the main module that will be imported in the other application.
```TS
export * from './app/main.module';
```

And this is our `package.json` file
```JSON
"scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build",
    "build:lib": "ng-packagr",
    "test": "ng test",
    "lint": "ng lint",
    "e2e": "ng e2e"
  },
"dependencies": {
    "@angular/animations": "^4.1.3",
    "@angular/common": "^4.1.3",
    "@angular/compiler": "^4.1.3",
    "@angular/core": "^4.1.3",
    "@angular/forms": "^4.1.3",
    "@angular/http": "^4.1.3",
    "@angular/platform-browser": "^4.1.3",
    "@angular/platform-browser-dynamic": "^4.1.3",
    "@angular/router": "^4.1.3",
...
}
```

When I launch `npm run build:lib`, I face into this error 
```
Building Angular library from C:\_Git\SFK\src\Sfk.WebComponents\ng-package.json
fs.js:583
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open 'C:\_Git\SFK\src\Sfk.WebComponents\node_modules\@angular\animations\src\my-component-tpl.html'
    at Object.fs.openSync (fs.js:583:18)
    at Object.fs.readFileSync (fs.js:490:33)
    at getFileData (C:\_Git\SFK\src\Sfk.WebComponents\node_modules\gulp-inline-ng2-template\parser.js:284:17)
    at C:\_Git\SFK\src\Sfk.WebComponents\node_modules\gulp-inline-ng2-template\parser.js:248:22
    at C:\_Git\SFK\src\Sfk.WebComponents\node_modules\async\dist\async.js:3830:24
[...]

PR #60, opened 2017-06-30
removed node_modules, excluded typings

davidenke:
As mentioned in #51 by @ChrisProlls in his [comment](https://github.com/dherges/ng-packagr/issues/51#issuecomment-307148324), we're unable to specify a source folder apart from the package.json file.
So we have to exclude the node_modules folder and the typings from the template glob.

### DIFF
--- a/lib/steps/assets.ts
+++ b/lib/steps/assets.ts
@@ -26,7 +26,7 @@ export const processAssets = (src: string, dest: string): Promise<any> => {
   return new Promise((resolve, reject) => {
     debug(`processAssets ${src} to ${dest}`);
 
-    vfs.src(`${src}/**/*.ts`)
+    vfs.src([`${src}/**/*.ts`, '!node_modules/**/*', '!**/*.d.ts'])
       .pipe(inlineNg2Template({
         base: `${src}`,
         useRelativePaths: true,
--- a/lib/steps/assets.ts
+++ b/lib/steps/assets.ts
@@ -26,7 +26,7 @@ export const processAssets = (src: string, dest: string): Promise<any> => {
   return new Promise((resolve, reject) => {
     debug(`processAssets ${src} to ${dest}`);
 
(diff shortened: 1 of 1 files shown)


dherges:
It breaks another build we have here in place, where we add a self-written `*.d.ts` for a 3rd party lib to the build (w/ triple-slash reference). Error is:

```
BUILD ERROR
Error at <root>/dist/.ng_build/@foo/bar/ts/lib/feature/feature-adapter
.service.ts:1:22: File '<root>/dist/.ng_build/@foo/bar/ts/lib/feature/feature.d.ts' not found.
```

I can live with excluding `node_modules` and adding a test for custom `*.d.ts` w/ triple-slash reference.

davidenke:
@dherges I excluded the definition files explictly because of an error showing up when the component definition is processed (Something about parsing template url of type `templateUrl: string`). So we could only exlude this very special definition?

davidenke:
@dherges Ngah, sorry this makes no sense at all. Your definition is not in node_modules, so why is it failing though?

dherges:
@davidenke If you exclude only `node_modules` (as you wrote before) does that solve your problem?

Because `*.d.ts` referenced through triple-slash should be copied imho. It's a bit hacky, because `procesAssets()` not only inlines and compiles scss, it also copies all the typescript sources to the intermediate directory. So it is actually intended to copy `*.d.ts` files that live inside the project directory.

davidenke:
@dherges Yes, it solves the problems with the `gulp-inline-ng2-template` plugin.

dherges:
Good! Then we have a solution.

I'll look at the issue later this evening or early tomorrow.

- reproduce the d.ts triple-slash reference in integrations
- apply your suggested test
- test and build all the libs :)
- publish new version


Cheers, David

ChrisProlls:
Thanks !!

Issue #66, opened 2017-07-07
sass imports from node_modules


davidenke:
My libraries mostly contain scss files which are consumed by components in the library itself and by projects depending on the library.

So sass imports pointing to `node_modules` dependencies can't be consumed by external styles since the paths have to be relative.

Angular CLI projects are using webpack by default, where those dependencies are imported by using the trailing tilde `~` pointing to the `node_modules` folder of the current project. To adopt this behavior we could use an importer function for the `node-sass` library:

> lib/steps/assets.ts:~70
```typescript
const sassImporter = (url: string): any => {
  if (url[0] === '~') {
    url = path.resolve('node_modules', url.substr(1));
  }

  return { file: url };
}

const pickRenderer = (filePath: string, ext: string[], file: string): Promise<string> => {

  switch (path.extname(filePath)) {

    case '.scss':
    case '.sass':
      debug(`rendering sass for ${filePath}`);
      return renderSass({ file: filePath, importer: sassImporter });

    case '.css':
    default:
      return Promise.resolve(file);
  }

}
```
@dherges I could pr this, if considered useful?

davidenke:
Closing this, since the PR has been merged :)

PR #67, opened 2017-07-07
Feature/scss node modules import

davidenke:
As described in #66 usages of tilde `~` in scss imports are dissolved to `node_modules`.


dherges:
@davidenke thanks!

I will check tomorrow. From first view, looks good!

dherges:
@davidenke yep, looks good! didn't also break other builds that I tested against!

Thanks for the PR!

davidenke:
@dherges very nice, thank you!

PR #70, opened 2017-07-11
use pkg name when `@<scope>` is undefined.

aitboudad:


dherges:
@aitboudad thanks! looks good!

Issue #73, opened 2017-07-14
Support rxjs operators w/ different import syntaxes


davidenke:
**Building with rxjs version 5.4.2 throws an error.**
Started build with `ng-packagr -p ./ng-package.json`.

Console output (shortened, removed usual npm error message:
```bash
Building Angular library from ./ng-package.json

BUILD ERROR
'merge' is not exported by node_modules/rxjs/observable/merge.js
Error: 'merge' is not exported by node_modules/rxjs/observable/merge.js
...
```

In the moment I have no idea why this happens, since Observable.merge hasn't been removed and is still exported properly at first sight...

dherges:
@davidenke can you reproduce this?

I think I've updated a few of "our's" libraries today to latest versions of Angular and RxJS and they still build fine. Our CI's went through all fine and I think it went up to typescript 2.4 and rxjs 5.4.x - there was a build error with rxjs and typescript 2.4 which was resolved in the latest rxjs version?!?

davidenke:
@dherges you're right, it turned out that I had issues with lodash. Btw. have you ever experienced issues with building relating lodash imports?

davidenke:
Okay, I discovered this issue again and built a simple test case.
Presuppose the following file structure:
```
.
+-- ng-package.json
+-- package.json
+-- public_api.ts
+-- tsconfig.json
```

Content of `ng-package.json`:
```json
{
  "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
  "src": ".",
  "workingDirectory": "./.packagr-cache",
  "lib": {
    "entryFile": "./public_api.ts",
    "externals": {}
  }
}
```

Content of `package.json`:
```json
{
  "name": "@test/test",
  "version": "0.0.0",
  "description": "",
  "scripts": {
    "build": "ng-packagr -p ./ng-package.json"
  },
  "author": "David Enke <post@davidenke.de>",
  "peerDependencies": {
    "@angular/animations": "^4.0.0",
    "@angular/cdk": "2.0.0-beta.8",
    "@angular/common": "^4.0.0",
    "@angular/core": "^4.0.0",
    "@angular/material": "2.0.0-beta.8",
    "rxjs": "^5.4.2"
  },
  "devDependencies": {
    "@angular/animations": "^4.0.0",
    "@angular/cdk": "2.0.0-beta.8",
    "@angular/common": "^4.0.0",
    "@angular/compiler": "^4.0.0",
    "@angular/core": "^4.0.0",
[...]

Issue #62, opened 2017-07-01
Support commonjs dependencies, e.g. class-transformer


sforner405:
We use ng-packagr for packaging a Angular 4 Module. 
Unfortunately it does not work with the import of class-transformer. It states the nodemodule/class-transfomer does not define / export plainclass. 
If we use EmptyObservable from Rxjs the same problem occurs. 
Is there a solution for it?

Thanks.

sforner405:
A link to the class-transfomer project. 
https://github.com/pleerock/class-transformer

dherges:
What exactly is not working?

sforner405:
We get the following error:
`'classToPlain' is not exported by node_modules/class-transformer/index.js
Error: 'classToPlain' is not exported by node_modules/class-transformer/index.js
`

Import in the service is:
`import { classToPlain, plainToClass } from "class-transformer";`

and the ng-package.json :
`{
    "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
    "src": "src",
    "dest": "dist",
    "workingDirectory": ".ng_build",
    "lib": {
        "entryFile": "app/core/core.module.ts"
    }
}`

dherges:
Can you please provide more context?

Can you run the build with `DEBUG=true` env variable? E.g. in pcakge.json:

```json
  "scripts": {
    "build": "DEBUG=true ng-packagr"
  }
```

Is the issue caused by rollup? If yes, is this an issue with `class-transformer` because it does not support ES2015 modules?

sforner405:
We get the following debug message:

[debug] rimraf /home/project_wizard_ui/dist
[debug] rimraf /home/project_wizard_ui/.ng_build
[debug] processAssets /home/project_wizard_ui/src to /home/project_wizard_ui/.ng_build/ts
[debug] render stylesheet /home/project_wizard_ui/src/app/app.component.css
[debug] postcss with autoprefixer for /home/project_wizard_ui/src/app/app.component.css
[debug] render stylesheet /home/project_wizard_ui/src/app/project-steps-activation/project-steps-activation.component.scss
[debug] rendering sass for /home/project_wizard_ui/src/app/project-steps-activation/project-steps-activation.component.scss
[debug] postcss with autoprefixer for /home/project_wizard_ui/src/app/project-steps-activation/project-steps-activation.component.scss
[debug] render stylesheet /home/project_wizard_ui/src/app/project-steps-partner/project-steps-partner.component.scss
[...]

PR #76, opened 2017-07-20
feat: use rollup-plugin-commonjs

DDeis:
As evoqued in #73 and #62


DDeis:
Include `node_modules/rxjs/**` only and add the possibility to include other commonjs dependencies manually with `ng-package.json` to #62 problem.

dherges:
Hi @DDeis,

I really like this improvement!

I suggest the following: we have integration testing samples (in the folder `integration` obvious). Add a test (a service or component or whatever) to the `src` that
 - a) uses the `import { map } from 'rxjs/operator/map';` syntax
 - b) uses the "old-fashioned"(?) `import 'rxjs/add/operator/map;` syntax

Then, the `angular.service.ts` used to demonstrate usage of the `lib.externals` property for the syntax in a. If possible, this test should be kept by another kind of import.

If you like, you can also create another sample, e.g. `integration/sample_rxjs` that demonstrates the issues you had (the motivation for the change) and the two syntax-ways of using rxjs operators (see above: a, b)

DDeis:
Hi @dherges, thank you for your feedbacks, I can't work on this weekend but I will update the PR follow ONG your recommandations as soon as I can and will check if 'rxjs/add/operator/*' Stills works.

Also I wanted to know what was your opinion about the knew 'commonjs' property in 'ng-package', do you think it's a proper name ? Do you think we we can include this property in 'externals' and create a new property inside for "old fashionned" symbols to have both ways to include together ?

DDeis:
Added sample to demonstrate the two syntax-ways of using rxjs operators as suggested.

dherges:
@DDeis Hi, thanks for your effort! I will look at it.

dherges:
@DDeis I made some smaller changes (tab indentation, update the description in the json schema).

However, when trying out this version in a larger project, the build errors:


```bash
BUILD ERROR
'BehaviorSubject' is not exported by ..\..\node_modules\rxjs\BehaviorSubject.js
Error: 'BehaviorSubject' is not exported by ..\..\node_modules\rxjs\BehaviorSubject.js
    at error (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:177:12)
    at Module.error$1 [as error] (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:8434:2)
    at Module.trace (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:8536:9)
    at ModuleScope.findDeclaration (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:8114:22)
    at Scope.findDeclaration (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:5907:39)
    at Node.bind (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:7165:29)
    at ....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:5701:50
    at Node.eachChild (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:5718:5)
    at Node.bind (....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:5701:7)
    at ....\node_modules\ng-packagr\node_modules\rollup\dist\rollup.js:5701:50

error Command failed with exit code 111.
```

lib/steps/rollup.ts
```diff
@@ -27,151 +29,26 @@ export const rollup = (opts: RollupOptions) => {
...
+    // external symbols passed from the user's ng-package.json
     ...opts.externals,
   };
 
+  const ROLLUP_COMMONJS_INCLUDE = [
+    // RxJS dependencies
+    'node_modules/rxjs/**',
+    // commonjs symbols passed from the user's ng-package.json
+    ...opts.commonjs
+  ];
```
dherges: `rollup-plugin-commonjs` does have an issue w/ monorepos.

```bash
$ cd packages/my-package
$ ../../node_modules/.bin/ng-packagr
```

Now, rollup will resolve the dependency to `../../node_modules/rxjs` but the value in `ROLLUP_COMMONJS_INCLUDE ` does not match.

In result, together w/ removing the default `externals` breaks existing builds.

DDeis:
Update the PR to allow specifying the path of the `node_modules` directory.
Would this solve the issue with monorepos ?

dherges:
That still breaks existing builds w/ their current configuration. So the change is requiring people to update their config in a project where they were building rxjs dependencies just fine!

Altogether, there are two things:
 - rxjs and their two import styles #73 
 - commonjs support #62 

Is it possible to separate the changes for these two? One for the rxjs thing, one for the commonjs.

DDeis:
Right, maybe we should have two PRs to address these issues distinctively:
- rxjs and their two import styles #73: adding missing rxjs dependencies
- commonjs support #62: using rollup-plugin-commonjs

DDeis:
PR splited into: 
- PR #82: rxjs two import styles #73
- PR #83: commonjs support #62

PR #83, opened 2017-07-27
feat: commonjs support

### LINKED ISSUE #62: Support commonjs dependencies, e.g. class-transformer

DDeis:
As evoqued in PR #76, splitting the PR.

This adds `rollup-plugin-commonjs` (resolves #62). 

By default it bundles CommonJS dependencies in `node_modules` and other non-CommonJS dependencies are ignored. Though, adds the possibility to include and exclude dependencies in `ng-package.json`:

```json
{
  "$schema": "../../lib/ng-package.schema.json",
  "lib": {
    "entryFile": "public_api.ts",
    "commonjs": {
      "include": ["node_modules/dependency_to_include/**", "..."],
      "exclude": ["node_modules/dependency_to_exclude/**", "..."]
    }
  }
}

```

Integration testing sample in `sample_commonjs` that demonstrate the use of a CommonJS dependency, e.g. `class-transformer`.

_Note: haven't tested with monorepos._


dherges:
Hm, I am able to built a larger, existing project w/ the change. So it's basically fine.

I have another question, though: the generated `commonjs.es5.js` (and the other bundles) include the whole `class-transformer` library. Is this inteded?

DDeis:
Changed `commonjs` property in `ng-package.json` to use it the same way as `rollup-plugin-commonjs`. This way we can use `namedExports`, e.g.: 
```json
{
  "$schema": "../../lib/ng-package.schema.json",
  "lib": {
    "entryFile": "public_api.ts",
    "commonjs": {
      "include": [
        "node_modules/class-transformer/**",
        "node_modules/lodash/**",
        "node_modules/moment/**"
      ],
      "namedExports": {
        "node_modules/lodash/index.js": ["identity"],
        "node_modules/lodash/moment.js": ["moment"]
      }
    }
  }
}
```

Demonstrate use of `lodash` in `sample_commonjs`. 

Tried to demonstrate the use of `moment.js` but it results in either `Cannot call a namespace ('moment')` when importing moment with `import * as moment from 'moment';` or  `"moment" has no default export` when importing with `import moment from 'moment';`. This [stackoverflow answer](https://stackoverflow.com/questions/39519823/using-rollup-for-angular-2s-aot-compiler-and-importing-moment-js/40346098#40346098) hint how to resolve this but I couldn't make it works...

As for the whole `class-transformer` library being included in the bundle, I think that the way `rollup-plugin-commonjs` works... I searched a way to only embed only what was needed (e.g. `serialize` in the sample), but found nothing :/

dherges:
There should be nothing embedded imo and `class-transformer` added to `peerDependencies`. Then, apps include `class-transformer` in their `dependencies` and that's it.

That's the way how it works for angular and all other dependencies since we don't want to ship a library multiple times.

DDeis:
In [rollup-plugin-commonjs readme](https://github.com/rollup/rollup-plugin-commonjs/blob/master/README.md) it states that rollup-plugin-commonjs:

> Convert CommonJS modules to ES6, so they can be **included** in a Rollup bundle

If I understand well that's not what we want and maybe we got it wrong about `rollup-plugin-commonjs` from the start?

So if we do not want anything to be embedded then we should add commonjs dependencies to rollup [external option](https://rollupjs.org/#big-list-of-options).

In fact, I forked ng-packaged to test it using `class-transformer` and `lodash` and it works fine with the current version of ng-packagr :smile: 
I struggled a bit and to make it works we need to add `"class-transformer/index": "class-transformer"` to `externals` in `ng-package.json`. (See https://github.com/DDeis/ng-packaged-commonjs-test)

PR #99, opened 2017-08-09
Implements reading excludes from project tsconfig

aminpaks:
## I'm submitting a...
- [x] Feature

## Description
This PR improves the process of build by including some configs from default `tsconfig.json` of the project.
For now I just implemented `exclude` but we may merge any other items.
In some project there are demos or other stuff that should not be included in `dist` folder after building the library. By this feature we can achieve this matter.

I also added a new launch to debug the project easier with the experiment debug feature of chrome.
Did some cleaning up and stuff that they're not really worth mentioning.

- [x] No

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,6 +6,20 @@
       "request": "launch",
       "name": "yarn samples:dev",
       "program": "${workspaceRoot}/integration/samples.js"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug samples:dev",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": [
+        "run-script",
+        "debug:samples:dev"
+      ],
+      "port": 9229,
+      "stopOnEntry": false,
+      "cwd": "${workspaceRoot}",
+      "protocol": "inspector"
     }
   ],
   "compounds": []
--- a/integration/samples/core/not-include/simple.ts
+++ b/integration/samples/core/not-include/simple.ts
@@ -1,0 +1,3 @@
+export interface UnusedInterface {
+  test: boolean;
+}
--- a/integration/samples/core/tsconfig.json
+++ b/integration/samples/core/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "compilerOptions": {
+    "baseUrl": "."
+  },
+  "exclude": [
+    "not-include"
+  ]
+}
--- a/integration/samples/typings/tsconfig.json
+++ b/integration/samples/typings/tsconfig.json
@@ -1,0 +1,5 @@
+{
+  "include": [
+    "excludes"
+  ]
+}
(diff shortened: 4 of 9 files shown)


dherges:
Hi @aminpaks,

oh cool! If i understand correctly a project-specific tsconfig is optional?

Initially, I had some [thoughts on what to do with configurations for tsc/ngc](https://github.com/dherges/ng-packagr/blob/master/DESIGN.md#ngc-tsconfigjson) ... I am a little bit in two minds: on the one side, it's good to add a project-specific config by the user. On the other side, there are some configurations which need to be set by the tool and cannot be given by the user.

What do you think?

aminpaks:
Hi @dherges,

To me it's better as much as possible to give the developer more options, but we can check on top of all of these to not break the build process. For example we can parse the configs and merge those that are correct and needed. One of these is `exclude` which I already proposed and a good one is`strickNullCheck` or `noImplicitAny` which are very important to build a `aot` compatible lib. `typescript` has a function to parse tsconfigs which would me my option to go with.
At the end there are only few items that `ng-packagr` cares to replace to get things done.

dherges:
That's the point.

When I decided for the opinionated approach, I thought it's better to provide reasonable, commonly accepted default configs rather than giving too much configuration options.

When this project started, it was all about "how can we reduce the amount of custom configuration?"

You are making a good point and the configuration option should be available. _What_ you are saying, I agree. I am not convinced about about the _How?_

I'd like to think about it again ... right now, I prefer two things:

- provide a default configuration in the `tsconfig.ngc.json`, if it's reasonable and a well-accepted best practice (`noImplicitAny`, `strictNullCheck` ?)
- add a custom `exclude` property to `ng-package.schema.json`, if the developers should be able to configure and customize the setting

You see that the rollup config is build from the `lib.externals` property. With rollup, we don't expose the configuration file directly ... I think we should keep it the same with the tsconfig.

aminpaks:
The part that I disagree is duplicating the tsconfig, if developer has enabled all possible/suitable options in their root `tsconfig.json` why we create another file? I don't think it's intuitive. Externals is another story, and won't fit in this discussion.
I strongly suggest to keep things as native as possible. If you wanna make things easier for developer add a new item in `ng-package.json` then merge `tsconfig.ngc.json` with that one and at the end make sure that main options for ngc are properly set.

dherges:
Can we find a reasonable convention for the default `tsconfig.ngc.json` in ng-packagr?

Are `noImplicitAny` and `strictNullChecks` recommended for Angular? I have seen usages of the two ... I have seen projects not using it. Do you have any resource (a talk from a conference, a blog post, ...) about this?

dherges:
I quickly skipped through [Jasen Aden's video](https://www.youtube.com/watch?v=unICbsPGFIA&feature=youtu.be). He's not using it. Angular CLI does it neither in their [tsconfig](https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/blueprints/ng/files/tsconfig.json)

aminpaks:
Apparently you don't get the point. `noImplicitAny` is necessary for AOT compilation.

PR #121, opened 2017-09-12
fix: add rollup commonjs plugin to support all library types

msusur:
## I'm submitting a...

- [X] Bug Fix
- [ ] Feature

## Checklist

- [X] Commit Messages follow the [Conventional Commits](https://conventionalcommits.org/) pattern, see current [`.vcmrc`](https://github.com/dherges/ng-packagr/blob/master/.vcmrc)

## Description

This PR is to fix an issue that is described in the [issue #227](https://github.com/rollup/rollup-plugin-commonjs/issues/227).

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "read-file": "^0.2.0",
     "rimraf": "^2.6.1",
     "rollup": "^0.49.0",
+    "rollup-plugin-commonjs": "^8.2.1",
     "rollup-plugin-node-resolve": "^3.0.0",
     "sorcery": "^0.10.0",
     "stylus": "^0.54.5",
--- a/src/lib/steps/rollup.ts
+++ b/src/lib/steps/rollup.ts
@@ -1,5 +1,6 @@
 const __rollup = require('rollup');
 const nodeResolve = require('rollup-plugin-node-resolve');
+const commonJs = require('rollup-plugin-commonjs');
 import { debug } from '../util/log';
 import { ROLLUP_GLOBALS } from '../conf/rollup.globals';
 
@@ -31,6 +32,7 @@ export const rollup = (opts: RollupOptions) => {
     input: opts.entry,
     plugins: [
         nodeResolve({ jsnext: true, module: true }),
+        commonJs(),
     ],
     onwarn: (warning) => {
         if (warning.code === 'THIS_IS_UNDEFINED') {
--- a/yarn.lock
+++ b/yarn.lock
@@ -71,6 +71,10 @@ abbrev@1:
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"
 
+acorn@^5.1.1:
+  version "5.1.2"
+  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.1.2.tgz#911cb53e036807cf0fa778dc5d370fbd864246d7"
+
 ajv@^4.9.1:
   version "4.11.8"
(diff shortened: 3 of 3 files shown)


dherges:
Hi Mert,

thank you for your effort!

Changes look good. It seems to not break existing builds (that's a good thing! :+1: 

Have you tested the changes with a commonjs dependency? I can't see any direct test case in the samples so that's why I am asking.

msusur:
Yes, it works perfectly with [angular2-moment](https://github.com/urish/angular2-moment) whereas wasn't working before the change. This change helps automatically resolve the external dependencies for commonjs, otherwise you need to resolve it using the config. To be honest it isn't a big deal, but I think it would be nice to have this as a built in feature. What do you think @dherges ?

Workaround would be

```json
{
  "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
  "lib": {
    "entryFile": "src/public_api.ts",
    "externals": {
      "angular2-moment": "angular2-moment/module"
    }
  }
}
```

dherges:
Checked that some existing builds do not break with current version. So I think it's good to go!

dherges:
I squash this as "feat: <..>" to master ... maybe, technically it could also be a fix ... I don't know tbh and it feels a little bit more like a new feature to me???

msusur:
Well I think it makes sense :)  @dherges 🍻

dherges:
Fuck it... I found one use case where we broke an existing build (= regression caused)

Prior to `rollup-plugin-commonjs`,  it was possible to embed a bundled dependency and it would register to the global scope. Example:

source
```ts
/// <reference path="./keycloak.d.ts" />
import "keycloak-js";
```

Resulting bundle ("foo.es5.js"):

```js
(function (window, undefined) {
/* ... */
    if (typeof module === "object" && module && typeof module.exports === "object") {
        module.exports = Keycloak;
    }
    else {
        window.Keycloak = Keycloak;
        if (typeof define === "function" && define.amd) {
            define("keycloak", [], function () { return Keycloak; });
        }
    }
})(window);
```

Result: `window.Keycloak` is defined. It is intentionally wanted to embed that IIFE `(function() { })(window)` because a wrapper/adapter around a 3rd party lib is created....


Now, with `rollup-plugin-commonjs`:

```js
function createCommonjsModule(fn, module) {
    return module = { exports: {} }, fn(module, module.exports), module.exports;
}
var keycloak = createCommonjsModule(function (module) {
    (function (window, undefined) {
/* ... */

        if ('object' === "object" && module && 'object' === "object") {
            module.exports = Keycloak;
        }
        else {
            window.Keycloak = Keycloak;
            if (typeof undefined === "function" && undefined.amd) {
                undefined("keycloak", [], function () { return Keycloak; });
            }
        }

});
```

Result: `window.Keycloak` is undefined since "module.exports"  is passed to the bundled dependency.

Users were (of course!) relying on `window.Keycloak` and `window['Keycloak']`.

So question is: what is the TypeScript import syntax to obtain a reference to exported symbols of the bundled library?!?

dherges:
Ok, the migration/solution in that very specific case is:

```ts
/// <reference path="./keycloak.d.ts" />
import "keycloak-js";
import * as Keycloak from 'keycloak-js'; // <-- added this import

const foo = new Keycloak();
```

Prior, the `new Keycloak()` call used to work because of the global-scoped stuff and `keycloak.d.ts` declaring that global variable....

msusur:
Do you think there is a way to automate those tests? We could see that happening when I first created the PR, and saved you a lot of time :)

dherges:
Hi @msusur,

it's possible to add an integration test. Either in one of the [existing samples](https://github.com/dherges/ng-packagr/tree/master/integration/samples) or add a new one.

Issue #167, opened 2017-10-07
Requesting alternate ng-packagr configuration capabilities


DavidParks8:
## Type of Issue

- [x] Feature Request

## Description

I would like to be able to configure ng-packagr without the need for an extra `ng-package.json` file. Specifically, I'd like to reuse an existing file such as `tsconfig.json` and/or `package.json`.  As more and more tooling packages get added, a repo becomes polluted with large amounts of config files that make it harder to onboard new developers. Ideally, I'd like new developers not have to know about tooling so they can focus on implementing new business logic.

dherges:
Yes, why not have multiple configuration locations.

1. `ng-package.json`
2. `package.json` with `ngPackage: {}` section
3. `tsconfig.json` with `ngPackage: {}` section

DavidParks8:
Implemented and merged.

PR #169, opened 2017-10-09
Adding alternate configuration sources

DavidParks8:
## I'm submitting a...

- [x] Feature

## Checklist

- [x] Test Included

## Description

Adding multiple configuration sources. The new behavior is first to read from the command line ng-package file, then read from the default `ng-package.json` file (if different from the command line file), then read from the `ngPackage` field from `package.json`. All configuration sources are merged using lodash in that order. This should lay the groundwork for extending multiple configs, should we want to. The primary motivation for this feature is #167.

- [x] No -> Care was taken to ensure the behavior is the same as before, if the feature is not used.


DavidParks8:
I did most of the feature work, but something isn't quite right that's causing the tests to fail. At first glance, I don't see anything obviously wrong. @dherges, maybe you know what's broken.

dherges:
Hmpf, first, can you try to `git rebase` or `git reset` your branch to the current `master` of this repo? Then, try to commit your changes in one commit and push it to the `master` of your fork. 

The build error on circle is:

```
BUILD ERROR
ENOENT: no such file or directory, open '/home/ubuntu/ng-packagr/dist/lib/conf/tsconfig.ngc.json'
Error: ENOENT: no such file or directory, open '/home/ubuntu/ng-packagr/dist/lib/conf/tsconfig.ngc.json'

module.js:473
      throw err;
      ^

Error: Cannot find module '/home/ubuntu/ng-packagr/dist/cli/ng-packagr.js'
    at Function.Module._resolveFilename (module.js:527:15)
      ^
```

So maybe a relative import path is wrong? Or the `process.cwd()` somehow stopped things from working?

dherges:
`ng-packagr/dist/lib/conf/tsconfig.ngc.json` ... is something in the typescript compiler wrong? During the `yarn build`? Because the [`postbuild` should copy that json file to the `dist` folder](https://github.com/dherges/ng-packagr/blob/master/package.json#L79-L81)

DavidParks8:
Please don't merge yet - I'm adding a test.

DavidParks8:
Ready for review.

integration/samples/package-json/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "A sample library with configuration in package.json",
+  "version": "1.0.0-pre.0",
+  "private": true,
+  "repository": "https://github.com/dherges/ng-packagr.git",
+  "peerDependencies": {
+    "@angular/core": "^4.1.2",
+    "@angular/common": "^4.1.2"
+  },
+  "ngPackage": {
+    "$schema": "../../../src/lib/ng-package.schema.json",
```
dherges: Ah, do you know any possibility to extend the default schema of a `package.json`? 

Here is one ... quite old from 2013: https://www.npmjs.com/package/package.json-schema

Which schema is used for example by vscode, atom, ...? Could think of extending one of those ...

dherges: Here is one: http://json.schemastore.org/package

dherges:
Double-checked the changes on a larger project and everything still :green_apple:  !

The only thing is the `$schema` for the json editing support in IDE. How can we get it to work?

schema

integration/samples/package-json/package.json
dherges: Got it!

Issue #116, opened 2017-09-03
Read `NgPackageConfig` from CLI arguments


Toxicable:
## Type of Issue
- [x] Feature Request

Right now having a ng-package.json appears redundant.
I propose that we should optionally be able to configure via the CLI both the entry file and output name.
for example 
`ng-packagr --entryFile public_api.ts --libName myLibName`
I have a situation where all I want to use as config is a command line command, no extra files should be needed

alan-agius4:
Thanks for this feature request, however it seems that there is not much interest in such feature from the community. Hence, I am going to close this.

Let’s revisit in the future if things change.

Issue #23, opened 2017-06-02
Write artefacts to harmonized `dist` folder layout independent of `entryFile` location


dherges:
@angular/core folder layout:

```
- public_api.ts
- src/
   |- core.module.ts
   |- foo.ts
   |- ...
```

other folder layout:

```
src/
  |- public_api.ts
  |- ui-lib.module.ts
  |- foo.ts
```

Support that both file layouts result in a `<dist>/src` folder being created

about-code:
Hi once again a question: what's the matter with `public_api.ts`? Hasn't there been a convention for TS-Libs that a package's public API should be exported by `index.ts`? FWIW TypeScript module resolution has even been implemented to be able to automatically detect `index.ts` or `index.d.ts` files in NPM packages. I guess if there's yet another `public_api.ts`, `index.ts` would just re-export anything from `public_api.ts`. Therefore I wonder why angular community moves towards `public_api.ts`.

Is it just to indicate an intent and make people more aware of paying attention to what's private or public or is `index.ts` meant to export more than just `public_api.ts`? The latter wouldn't make sense to me, though.

dherges:
@about-code the `public_api.ts` "pattern" goes down to Angular's "Flat ECMAScript Modules" idea. It needs an entry file that references all symbols of a lib and then creates a so-called "flat-module file" (configured somewhere down the toolchain w/ `ngc`) for the imports. I think jason Aden talks about this: https://www.youtube.com/watch?v=unICbsPGFIA&feature=youtu.be

about-code:
David, thank you for the link. Jason gave a good talk which helped me packaging a lib prior to evaluating *ng-packagr*. However, I couldn't find a hint on why feeding a `public_api.ts` to `ngc` is beneficial over `index.ts`.

@jasonaden would you mind elaborate on what the reasons are to export a library's API via an extra `public_api.ts` instead of doing it via an `index.ts` on the root of a package? Is it a *proposed* pattern or an Angular-internal convention [Material2](https://github.com/angular/material2) seems to completely omit `index.ts` while [`@angular/core/index.ts`](https://github.com/angular/angular/blob/master/packages/core/index.ts) just re-exports `public_api.ts`.

[...]

PR #184, opened 2017-10-15
feat: dynamic secondary entry points

### LINKED ISSUE #107: Package secondary entrypoints (e.g. "@my/lib/testing")
#### Type of Issue

- [ ] Bug Report
- [x] Feature Request

#### Description

Angular Package Format "describes secondary entry points": https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview

We should add a possibility to compile and package such secondary entry points.

#### How To Reproduce

n/a

#### Expected Behaviour

TBD

Usage w/ ng-packagr must be further specified!

End result should be, say our already existing lib is `@my/lib`:

```ts
import { TestingStuff } from '@my/lib/testing';
```

#### Version Information

```
ng-packagr: v1.0.0-pre.14
```

### LINKED ISSUE #178: Async all the way?
## Type of Issue

```
[x] Refactor request
```

## Description

Any issue with me making everything `await`-based instead of `.then`-based?
Some benefits are summarized [here](https://www.twilio.com/blog/2015/10/asyncawait-the-hero-javascript-deserved.html).

DavidParks8:
## I'm submitting a...

```
[x] Feature
```

## Checklist

- [x] Tests for the changes have been added - there is a new sample project

## Description

Usually I like to make small targeted commits, but I needed a few of these features by next Tuesday and got a bit carried away. Please stick with me as I explain what's changed. This PR has a bit of fix #178, but it also might remove the need for #116 by allowing `yarn ng-packagr -p package.json`. It also should take care of #23 since `lib.entryFile` is now resolved by expanding it's path relative to the current `package.json`. All of those issues were not the focus of the changes, they just happened to get taken care of along the way. My actual goal was to fix #107 in a way that required minimal configuration and the least amount of ng-packagr overhead as possible. So this PR introduces dynamically discovered secondary entry points. No extra ngPackage field is required. Simply put `package.json` files in sub folders and ng-packagr does the rest. 

There are also a few extra goodies in the code such as the working directory getting deleted after the library is built, and each entry point now gets built in it's own isolated build folder and merged to dist at the end (so technically they could be built in parallel if you choose).

##### So how do I use it?

All you have to do is create a `package.json` file and put it where you want a secondary entry
point to be created. One way this can be done is by mimicking the folder structure of the 
following example which has a testing entry point in addition to it's main entry point.

```
src  
├── testing   
│   ├── *.ts   
│   ├── public_api.ts  
│   └── package.json  
├── *.ts  
├── public_api.ts  
├── ng-package.json   
└── package.json   
```

The contents of the secondary `package.json` can be as simple as:
```json
{}
```

No name is required. No version is required. It's all handled for you by ng-packagr!
When built, the secondary bundles would be accessible as `$(your-primary-package-name)/testing`.

You can change the entry point file by using the `ngPackage` configuration field in your secondary `package.json`.
For example, the following would use `index.ts` as the secondary entry point:

```json
{
  "ngPackage": {
    "lib": {
      "entryFile": "index.ts"
    }
  }
}
```

```
[x] No - the internals are different, but the external interface should be backwards compatible
```


src/cli/main.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as path from 'path';
+import { createNgPackage, NgPackagrCliArguments } from './../lib/ng-packagr';
+
+function parseProjectPath(parsed: string): string {
+  return parsed || path.resolve(process.cwd(), 'ng-package.json');
+}
+
+program
+  .name('ng-packagr')
+  .option('-p, --project <path>', 'Path to the \'ng-package.json\' or \'package.json\' file.', parseProjectPath)
```
dherges: The default value must come as the 4th argument to `.option(one, two, three, default)`.

Example:

```js
const DEFAULT_PROJECT_PATH = path.resolve(process.cwd(), 'ng-package.json');

program
    .name('ng-packagr')
    .option('-p, --project <path>', 'Path to the \'ng-package.json\' or \'package.json\' file.', parseProjectPath, DEFAULT_PROJECT_PATH)
    .parse(process.argv);
```

Change above changed the default handling for calling `ng-packagr` w/o arguments, thus causing existing builds to fail (e.g. `"scripts": { "build": "ng-packagr" }`

src/lib/steps/package.ts
```diff
@@ -28,105 +28,213 @@ const resolvePaths = (workingDirectory: string, packageConfig: NgPackageConfig):
...
-          // read 'ng-package.schema.json'
-          return schemaPromise.then((SchemaClass) => {
-            const schema = new SchemaClass(ngPkg);
+  let ngPackage: NgPackageConfig = await readNgPackageFile(ngPackageFile);
+  const packageJson = await readJson(packageJsonFile);
 
-            return new NgPackage(pkg, finalPackageConfig, dir, schema);
-          });
+  ngPackage = merge(ngPackage, packageJson.ngPackage, arrayMergeLogic);
+  ngPackage.lib.externals = rootPackage.libExternals;
```
dherges: Searching for ng-package config at `...\packages\components\lib\alert\ng-package.json`

```
Cannot set property 'externals' of undefined
TypeError: Cannot set property 'externals' of undefined
    at ...\node_modules\ng-packagr\lib\steps\package.js:166:33
```

src/lib/steps/package.ts
```diff
@@ -28,105 +28,213 @@ const resolvePaths = (workingDirectory: string, packageConfig: NgPackageConfig):
...
+async function readSecondaryPackage(rootPackage: NgPackageData, filePath: string): Promise<NgPackageData> {
+  const baseDirectory = path.dirname(filePath);
+  const ngPackageFile = path.resolve(baseDirectory, 'ng-package.json');
+  const packageJsonFile = path.resolve(baseDirectory, 'package.json');
 
-          // read 'ng-package.schema.json'
-          return schemaPromise.then((SchemaClass) => {
-            const schema = new SchemaClass(ngPkg);
+  let ngPackage: NgPackageConfig = await readNgPackageFile(ngPackageFile);
+  const packageJson = await readJson(packageJsonFile);
```
dherges: We should allow empty `package.json` files. I think some, maybe most, people, will create a file w/o contents and then run in the following error:

 ```
SyntaxError: ...\packages\components\lib\alert\package.json: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
```

(Maybe related to the other error ( `Cannot set property 'externals' of undefined`)

dherges: Another thinking: `new SchemaClass(ngPkg).$$get('...')` used to do default configuration values. As this got removed. It may have changed default value behaviour!

src/lib/steps/package.ts
DavidParks8: Fixed with a simple `if` statement.

src/cli/main.ts
DavidParks8: Fixed as requested

src/lib/steps/package.ts
DavidParks8: Added a function that returns `{}` instead of throwing when parsing json. I checked the schema and found that only one field was not matching the old default: `lib.entryFile`. I made a change to the root package's parsing so that it defaults to `src/public_api.ts` again, but I'm thinking that the secondary packages should default to `public_api.ts` so that users don't have to change their structure to allow an empty package.json. Since you are the owner, it is ultimately your decision, but I have seen many more secondary packages that do not have a second src folder, and it would make sense to cater to what we think most people will want.

DavidParks8: Something that just occurred to me: why does `ngPackage.src` default to `.` when `ngPackage.lib.entryFile` defaults to `src/public_api.ts`? Wouldn't it make more sense to have both fields either assume the existence of `/src`  or have both fields assume `.` is inside `/src'?

Rather, it seems odd that the defaults aren't (`.` & `public_api.ts`) or (`./src` & `src/public_api.ts`). Personally, I would prefer the former because that's how most of my libraries are structured.

dherges:
Ok! :+1: 

Tested secondary entrypoint w/ custom project. Secondary entrypoint is bundled. Existing builds still build. I ran into an error w/ an empty package.json file.

Built [clarity-angular](https://github.com/vmware/clarity/tree/master/src/clarity-angular) on master. Build still ok.

I'll merge this now and cut a pre-release for a `1.5.0-rc.0`. I.e.: you can switch to that new version and start using the new feature. Everybody with a `^1.x` version range will not (immediately) switch to the new version so it gives a little bit more time to test again. Because it's such a big change, I think this should be a good way!

PR #197, opened 2017-10-19
fix: produce correct secondary package paths

### LINKED ISSUE #191: Secondary entrypoints: package.json contains backslash on windows
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

Building a secondary entrypoint on a windows machine. Creates `package.json` for secondary:

```json
{
  "main": "..\\../bundles/components-lib\\alert.umd.js",
  "module": "..\\../@orbis-u/components/lib\\alert.es5.js",
  "es2015": "..\\../@orbis-u/components/lib\\alert.js",
  "typings": "alert.d.ts",
  "metadata": "alert.metadata.json",
  "name": "@orbis-u/components/lib\\alert"
}
```

#### How To Reproduce

Build on windows machine.

#### Expected Behaviour

Should contain forward slashes only (both for pathes and package name)

#### Version Information

```
ng-packagr: v1.5.0-rc.0
node: v8.4.0
```

DavidParks8:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description
These changes should hopefully fix #191. Give it a go and tell me if anything else pops up.

```
[ ] Yes
[x] No
```


src/lib/util/path.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// by exporting this way, we can add custom functions and still be type safe
+export const path = {
+  ...nodePath,
+  ensureUnixPath: (path?: string): string | null => {
+    if (!path) {
+      return null;
+    }
+
+    return path.replace(/\\/g, '/');
```
dherges: Hmpf. Ok. Isn't there a node API for reading or normalizing a platform-independent path?

If not, we can stick with the regex replace but I'd like to use [`path.sep`](https://nodejs.org/api/path.html#path_path_sep) then.

Wdyt?

src/lib/model/ng-artifacts-factory.ts
```diff
@@ -30,8 +31,8 @@ export class NgArtifactsFactory {
       main: this._unixPathJoin(rootPathFromSelf, 'bundles', `${this._makeUmdPackageName(ngPkg)}.umd.js`),
```
dherges: Is `rootPathFromSelf` guaranteed to have forward slashes `/`?

```
const rootPathFromSelf: string = path.relative(ngPkg.sourcePath, ngPkg.rootSourcePath);
```

I can only check this in the next week when being back at the office Windows machine.

src/lib/util/path.ts
DavidParks8: I tried the `path.posix.normalize` api and found that it seems to resolve things like `/a/../b` to be `/b`, but it didn't seem to make path separators uniform. Maybe I was using it wrong. Node isn't my main platform. I agree with the `path.sep` change if we keep the `.replace`.

src/lib/model/ng-artifacts-factory.ts
DavidParks8: It is not guaranteed to have forward slashes, but the `this._unixPathJoin` function should replace windows with posix path separators

dherges:
Tested on windows machine!  Works!

PR #268, opened 2017-11-10
fix: depend on user's typescript and tsc-wrapped

dherges:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Previously, a user of ng-packagr could install an incompatible typescript version for ng-packagr. Prevent inadvertent typescript installs by depending on a user's typescript isntallation (peerDependencies). Same for tsc-wrapped.

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "ng-packagr": "./cli/main.js"
   },
   "dependencies": {
-    "@angular/tsc-wrapped": "^4.4.5",
     "@ngtools/json-schema": "^1.1.0",
     "autoprefixer": "^7.1.1",
     "browserslist": "^2.1.5",
@@ -46,9 +45,12 @@
     "sorcery": "^0.10.0",
     "stylus": "^0.54.5",
     "ts-node": "^3.0.4",
-    "typescript": "^2.3.4",
     "uglify-js": "^3.0.7",
     "vinyl-fs": "^2.4.4"
+  },
+  "peerDependencies": {
+    "@angular/tsc-wrapped": "~4.4.5",
+    "typescript": "~2.3.4"
   },
   "devDependencies": {
     "@angular/cdk": "^2.0.0-beta.11",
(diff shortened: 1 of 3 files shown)


jeffora:
Hey, I know this is already merged, but I just wanted to mention that this change *is* a breaking change. In a CI scenario where a package depends on `"ng-packagr": "^1.0.0"` or `"ng-packagr": "~1.5.0"`, the build can break with no other code changes.

If `1.5.1` is installed, and the project including this doesn't include `@angular/tsc-wrapped`, `ng-packagr` will fail to run. The exact same project would have run with no issues at all on `1.5.0`.

dherges:
@jeffora Understand. For `typescript` this is not an issue because ~almost~ every Angular project will have typescript in their `devDependencies` and thus it was already installed to `node_modules/typescript`. `@angular/tsc-wrapped`, and the Angular core team said it was an internal package, was probably never hoisted to a direct child of `node_modules` but kept further down tree.... hmpf.

Sorry about that. I should've planned the release better by collecting the breaking changes for 2.x release candidates and incrementally improving that towards Angular 5 support.... :/

dherges:
- Create branch `v1` starting from tag `v1.5.1`
- revert commit 42b2f08
- re-apply change for `typescript` as `peerDependency`
- release `v1.5.2`
- cherry-pick 4068664 (feat) on branch `v1`
- release `v1.6.0` w/ Angular v4 support only

At that point, `v1.x` releases are finished.

Go back to `master`, manually copy&paste the changelog, manually release a `2.0.0-rc.0` (which needs to be npm-tagged `next`).

Does that sound good?

I'm curious why at vmware/clarity there were no issues as somehow tsc-wrapped has hoisted to node_modules for them?!?

jeffora:
Thanks for responding @dherges. That fix sounds viable to me, but we're also ok at the moment just pinning to 1.4.x.

Perhaps vmware/clarity had tsc-wrapped installed as a devDependency or similar in root projects?

Issue #251, opened 2017-11-06
Consider including scope name in moduleName


JoostK:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Currently, the moduleName of a library is set to the package's name without the scope name. I would like to see the possibility/default to include the scope name in the moduleName, as a package name such as `core` is meaningless without scope name and will likely cause conflicts.

I can whip up a PR if you feel this may be useful. I'm not quite sure if including the scope name by default is warranted as it would be a breaking change, although I feel like it is a more sensible default. What is your thought on how to go about this?

dherges:
Hi @JoostK,

yes, you're right. A good starting point for a PR is the `ngc.ts`:

https://github.com/dherges/ng-packagr/blob/c5c32c5/src/lib/steps/ngc.ts#L14-L15

PR #280, opened 2017-11-13
fix: include scope name in module name of Rollup bundle

JoostK:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

This PR changes the UMD module name of libraries to include the scope name. See #251 for background information on this change.

```
[x] Yes
[ ] No
```

Yes, this is a breaking change for users that have published libraries before. This change causes new builds to get assigned a different module name if a scope is being used.

### DIFF
--- a/integration/samples/core/specs/bundle.ts
+++ b/integration/samples/core/specs/bundle.ts
@@ -1,0 +1,23 @@
+import { expect } from 'chai';
+import * as fs from 'fs';
+import * as path from 'path';
+
+describe(`@sample/core`, () => {
+
+  describe(`bundle core.umd.js`, () => {
+    let BUNDLE;
+    before(() => {
+      BUNDLE = fs.readFileSync(
+        path.resolve(__dirname, '..', 'dist', 'bundles', 'core.umd.js'), 'utf-8');
+    });
+
+    it(`should exist`, () => {
+      expect(BUNDLE).to.be.ok;
+    });
+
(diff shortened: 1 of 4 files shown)


dherges:
Hi @JoostK,

cool, thank you! The changes look good to me!

For the "BREAKING CHANGES" ... hm ... how is it supposed to be? After the changes, you can still use ng-packagr the same way you did before. There's no change needed and it will still produce the binaries in Angular Package Format. However, for consumers of the library, they will now have a "BREAKING CHANGE" when the library authors publish a new version. So, technically, upgrading to ng-packagr@1.6.x will not break. But library authors would need to introduce a (potentially) BREAKING CHANGE to their changelog when they decide to use ng-packagr@1.6.x.

JoostK:
@dherges Absolutely right. I am being conservative with the indication that this is a breaking change, as for some users it may be considered that way. They cannot upgrade without introducing a breaking change to their library, which would warrant a major version increment (if following semver).

dherges:
Never mind. Things were messed up in v1.5.2 and v1.6.0-rc.0.

So I now released v1.5.2 and v1.6.0 with Angular v4 support (from a `v1` branch).

The `master` branch will continue towards Angular v5 and I'll try to publish a series of `v2.0.0-rc.x` for Angular v5. In these rc versions, let's collect breaking changes so I'll merge yours in.

PR #279, opened 2017-11-12
refactor: use TypeScript custom transformers for resource inlining

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
[x] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

- [ ] Commit Messages follow the [Conventional Commits](https://conventionalcommits.org/) pattern
  - A feature commit message is prefixed "feat:"
  - A bugfix commit message is prefixed "fix:"
- [ ] Tests for the changes have been added

## Description

See #261 

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #292, opened 2017-11-17
fix: compiler no longer throwing an error

crowmagnumb:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

complier-cil.performCompilation was now silently failing resulting in no compiled code getting written and so when the source mapper ran it got an error because the file it was trying to map didn't exist. Again because performCompilation was silently failing. This checks the result of the compilation to see if it resulted in an error and tries to give the best info it can about this and then throws an exception to stop the whole packaging process.  Notice I had problems getting the typescript namespace (ts) recognized in the ng-packagr build process so  I had to comment that out.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/steps/ngc.ts
+++ b/src/lib/steps/ngc.ts
@@ -1,9 +1,9 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import { performCompilation, readConfiguration, AngularCompilerOptions, Diagnostic } from '@angular/compiler-cli';
+import { performCompilation, readConfiguration, AngularCompilerOptions, exitCodeFromResult, formatDiagnostics } from '@angular/compiler-cli';
 import { NgPackageData } from '../model/ng-package-data';
 import { readJson, writeJson } from 'fs-extra';
-import { debug, error, warn, info } from '../util/log';
+import { debug } from '../util/log';
 
 
 async function prepareTsConfig(ngPkg: NgPackageData, outFile: string): Promise<void> {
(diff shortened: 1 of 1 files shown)


dherges:
Hi @crowmagnumb,

thanks for the PR!

Please bare with me and be a little patient, there is change coming in #279

src/lib/steps/ngc.ts
```diff
@@ -40,7 +44,32 @@ export async function ngc(ngPkg: NgPackageData, basePath: string): Promise<strin
 
   // invoke ngc programmatic API
   const compilerConfig = readConfiguration(tsConfigPath);
-  performCompilation(compilerConfig);
+  const compilerResult = performCompilation(compilerConfig);
+
+  if (!compilerResult.emitResult) {
```
dherges: Luckily, there are helper functions for this from `@angular/compiler-cli`.

```ts
  const exitCode = ng.exitCodeFromResult(result.diagnostics);
  if (exitCode === 0) {
    return Promise.resolve(
      path.resolve(ngPkg.sourcePath, tsConfig.options.outDir, tsConfig.options.flatModuleOutFile)
    );
  } else {
    return Promise.reject(new Error(
      ng.formatDiagnostics(result.diagnostics)
    ));
  }
```

If you like an early fix, I ask you to make a PR with these changes. Otherwise, let's wait for #279. I just need a few more days to correct the "write staged files to dist".

crowmagnumb: Well, I'm not in a super hurry on that front but I did make the changes you suggested and it's much better so I just added them here. :)

dherges:
Travis is not running for this PR?!? Why the heck that is... don't know ... Circle is green, and the change looks good to me, so let's go for it!

dherges:
@crowmagnumb Is now published in `2.0.0-rc.0` / `ng-packagr@next`

https://yarnpkg.com/en/package/ng-packagr

Issue #321, opened 2017-11-27
Using @ in package name


pl4yradam:
## Type of Issue

```
[ x ] Bug Report
[ ] Feature Request
```

## Description

When writing a package that contains an @ symbol, causes the project to not be publishable
Packge name: @something/something

#### How To Reproduce

Create a new package, with a name of @something/something, then go through npm run packagr, when the /dist folder is made you will notice the folder will have @something in it

When you try to do npm publish, the es5 file cannot be found (the path is looking for /something not /@something. 

#### Expected Behaviour

The correct file to be used.

_please describe what behaviour or result you expected_

#### Version Information

```
ng-packagr: v1.6.y
node: v9.x.y
@angular: v5.x.y
rxjs:
zone.js:
```

To fix, I have just removed the @ symbol from the dist folder

dherges:
Hi @pl4yradam,

what npm version do you have?

It sounds very much like  npm/npm#19222

dherges:
The issue will be circumvented by #319 and #322 as Package Format 5 drops the "@org" folder names

Issue #319, opened 2017-11-26
Angular package format has changed


DavidParks8:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Angular package format (link from readme) now shows package format version 5. Ng-Packagr should be updated to reflect this new format.

#### Version Information

```
ng-packagr: v2.0.0-rc.3
@angular: v5.x.y
```

DavidParks8:
I'm willing to implement this. @dherges should I go ahead before, or after #279?

alan-agius4:
#322

dherges:
Hi @alan-agius4 @DavidParks8 ,

what has changed beside the folder names for "esm5", "esm2015"?

Metadata JSON files carry a version flag. It has been increased from "version: 3" to "version: 4" by bumping `ngc` to Angular v5.

Are there more changes?

DavidParks8:
It seems that es5 files no longer have .es5.js and are just .js inside the es5 folder. That’s another change I see.

alan-agius4:
Ah yeah totally forgot about that. I doesn't have it any more. It's by the
new specs.

PR #322, opened 2017-11-27
feat: align distributed files with `Angular Package Format v5.0`

### LINKED ISSUE #257: package.json in output is compressed
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

package.json in output is compressed

#### How To Reproduce

use the [ng-packaged](https://github.com/dherges/ng-packaged) and install the latest ng-packagr to perform packaging.

#### Expected Behaviour

no compressed for package.json

#### Version Information

```
ng-packagr: v1.5.0
node: v8.9
@angular: v5.0.0
```

alan-agius4:
Fixes #257 #319  #321 

- FESM5 and FES2015 modules are not into seperate directories `/esm5` and `/esm2015` respectively.
- Secondary entry points output files are longer nested under
`/secondary-package`, only typings  are.

More info:
https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview

## I'm submitting a...

```
[ ] Bug Fix
[X] Feature
```

## Checklist

## Description
Align with Angular Package Format v5.0
https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview

```
[] Yes
[X] No
```

### DIFF
--- a/src/lib/bundler.ts
+++ b/src/lib/bundler.ts
@@ -20,9 +20,9 @@ import * as log from './util/log';
 export async function generateNgBundle(ngPkg: NgPackageData): Promise<void> {
 
   log.info(`Generating bundle for ${ngPkg.fullPackageName}`);
-  const artifactFactory: NgArtifactsFactory = new NgArtifactsFactory();
-  const baseBuildPath: string = `${ngPkg.buildDirectory}/ts${ngPkg.pathOffsetFromSourceRoot}`;
-  const artifactPaths: NgArtifacts = artifactFactory.calculateArtifactPathsForBuild(ngPkg);
+  const artifactFactory = new NgArtifactsFactory();
+  const baseBuildPath = `${ngPkg.buildDirectory}/ts${ngPkg.pathOffsetFromSourceRoot}`;
+  const artifactPaths = artifactFactory.calculateArtifactPathsForBuild(ngPkg);
 
   // 0. CLEAN BUILD DIRECTORY
   log.info('Cleaning bundle build directory');
@@ -34,9 +34,9 @@ export async function generateNgBundle(ngPkg: NgPackageData): Promise<void> {
 
   // 2. NGC
   log.info('Running ngc');
-  const es2015EntryFile: string = await ngc(ngPkg, baseBuildPath);
+  const es2015EntryFile = await ngc(ngPkg, baseBuildPath);
   // XX: see #46 - ngc only references to closure-annotated ES6 sources
-  await remapSourceMap(`${baseBuildPath}/${ngPkg.flatModuleFileName}.js`);
+  await remapSourceMap(es2015EntryFile);
 
   // 3. FESM15: ROLLUP
   log.info('Compiling to FESM15');
@@ -69,7 +69,7 @@ export async function generateNgBundle(ngPkg: NgPackageData): Promise<void> {
 
   // 6. UMD: Minify
   log.info('Minifying UMD bundle');
-  const minifiedFilePath: string = await minifyJsFile(artifactPaths.main);
+  const minifiedFilePath = await minifyJsFile(artifactPaths.main);
   await remapSourceMap(minifiedFilePath);
 
   // 7. SOURCEMAPS: RELOCATE ROOT PATHS
@@ -82,7 +82,7 @@ export async function generateNgBundle(ngPkg: NgPackageData): Promise<void> {
 
   // 9. WRITE PACKAGE.JSON and OTHER DOC FILES
   log.info('Writing package metadata');
-  const packageJsonArtifactPaths: NgArtifacts = artifactFactory.calculateArtifactPathsForPackageJson(ngPkg);
+  const packageJsonArtifactPaths = artifactFactory.calculateArtifactPathsForPackageJson(ngPkg);
   await writePackage(ngPkg, packageJsonArtifactPaths);
 
   log.success(`Built Angular bundle for ${ngPkg.fullPackageName}`);
--- a/src/lib/model/ng-artifacts-factory.ts
+++ b/src/lib/model/ng-artifacts-factory.ts
@@ -12,12 +12,13 @@ export class NgArtifactsFactory {
       return this._makeEsmPackageNameVal;
     }
 
-    const { packageNameWithoutScope, flatModuleFileName, buildDirectory } = ngPkg;
+    const { packageNameWithoutScope, flatModuleFileName, buildDirectory, pathOffsetFromSourceRoot } = ngPkg;
 
     if (packageNameWithoutScope === flatModuleFileName) {
       this._makeEsmPackageNameVal = `${flatModuleFileName}.js`;
     } else {
-      const modulePath = path.join(buildDirectory, 'esm2015', `${flatModuleFileName}.js`);
+      const pathFromRoot = buildDirectory.replace(pathOffsetFromSourceRoot, "");
+      const modulePath = path.join(pathFromRoot, 'esm2015', `${flatModuleFileName}.js`);
       const pkgName = pathExistsSync(modulePath) ? packageNameWithoutScope : flatModuleFileName;
       this._makeEsmPackageNameVal = `${pkgName}.js`;
     }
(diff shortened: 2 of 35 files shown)


dherges:
Hi  @alan-agius4 ,

thanks for the PR!

What has changed in Angular Package Format 5 beside the "esm5" / "esm2015" folder names?

I have been working on improving the way `tsc` / `ngc` are being used (see #279) and I'm afraid we'll run into merge conflicts. I hadn't had enough time to fix the output for the secondary entry points in #279 so it's still open.

Are the changes really breaking? If only the internal folder layout of the generated package has changed, it shouldn't be.

alan-agius4:
Well, I marked them as breaking change due to the fact, if a consumer of a library built with ng-packagr is using system-js they will need to update the entry points, that's the main reason, the breaking change is infect the package layout. Of course if you don't agree I can always amend the commit message.

The changes are as follows;
The esm5/esm2015/bundles folder layout

so basically the output will look like something like;
```
--esm2015
--esm5
--bundles
--src
--index.metadata.json
--index.d.ts
--index.ts
```

dherges:
Why a System.JS has to be changed? Shouldn't the path be managed by the property `"main": "bundles/foobar.umd.js"` in `package.json`? The UMD module ID isn't affected by the changed if I see right.

I am not using System.JS so I am asking beginner's questions... :-)

dherges:
Also, looking at the tests, the path for the UMD artefact hasn't changed, right?

alan-agius4:
Well I am not a big `System.JS` user but usually I tend to configure it this way

That said, I just realized that bundles layout are still the same.

```json
System.config({
	baseURL: "/base/",
	paths: {
		"npm:": "node_modules/"
	},
	map: {
		"@angular/animations": "npm:@angular/animations/bundles/animations.umd.js",
		"@angular/animations/browser": "npm:@angular/animations/bundles/animations-browser.umd.js",
      }
}
```

alan-agius4:
@dherges indeed the umd is unchanged, so I guess it's safe to remove the breaking change.

src/lib/model/ng-artifacts-factory.ts
```diff
@@ -13,26 +13,28 @@ export class NgArtifactsFactory {
...
 
     return {
-      main: path.join(ngPkg.buildDirectory,'bundles', this._makeUmdPackageName(ngPkg)),
-      module: path.join(ngPkg.buildDirectory, ngPkg.fullPackageName + '.es5.js'),
-      es2015: path.join(ngPkg.buildDirectory, ngPkg.fullPackageName + '.js'),
-      typings: path.join(pathFromRoot, ngPkg.flatModuleFileName + '.d.ts'),
-      metadata: path.join(pathFromRoot, ngPkg.flatModuleFileName + '.metadata.json')
+      main: path.join(buildDirectory, 'bundles', this._makeUmdPackageName(ngPkg)),
+      module: path.join(buildDirectory, 'esm5', `${flatModuleFileName}.js`),
+      es2015: path.join(buildDirectory, 'esm2015', `${flatModuleFileName}.js`),
```
dherges: Do we have potentiaal file names conflicts here?

https://github.com/alan-agius4/ng-packagr/blob/c6f6dfb8c2cd17f3c1803d6431b9494570983594/src/lib/model/ng-package-data.ts#L65-L67

What if `@foo/bar/testing`, `@foo/bar/sub/testing` both have the "flatModuleFileName": "testing"

alan-agius4: won't in that case `buildDirectory` be `@foo/bar/sub`?

so i assume it should be 

```
  --esm2015
  ----testing.js
  ----sub
  ------testing.js
```

alan-agius4: Yes indeed, I am thinking how we can address that.

dherges: Let's take the "full" package name excluding the npm org. E.g.:

`@foo/bar/sub/testing` should be `bar-sub-testing.js`
`@foo/bar/testing` should be `bar-testing.js`

alan-agius4: Agreed, unless you want to do like angular are doing something like;
ie; create a subfolder, with the library name if the file already exists.

image

dherges: Ok! Let's take the subfolder then!

alan-agius4: I did the change, let me know what you think. I still need to add some unit tests, but i'll do these later today or tomorrow :)

alan-agius4: Tests added

src/lib/model/ng-artifacts-factory.ts
```diff
@@ -17,7 +17,7 @@ export class NgArtifactsFactory {
     if (packageNameWithoutScope === flatModuleFileName) {
       this._makeEsmPackageNameVal = `${flatModuleFileName}.js`;
     } else {
-      const pathFromRoot = buildDirectory.replace(pathOffsetFromSourceRoot, "");
+      const pathFromRoot = buildDirectory.replace(pathOffsetFromSourceRoot, '');
```
alan-agius4: this is required because each package has it's own build directory, it can be removed, if everything is built in a single folder.

src/lib/steps/uglify.ts
```diff
@@ -38,7 +38,9 @@ export async function minifyJsFile(inputPath: string): Promise<string> {
     throw result.error;
   }
 
-  await writeFile(outputPath, result.code);
-  await writeFile(sourcemapOut, result.map);
+  await Promise.all([
```
DavidParks8: Per my comment about slowing the disk down, I'd like to see any and all parallel disk writes turned into synchronous disk flushes.

src/lib/model/ng-artifacts-factory.ts
```diff
@@ -1,38 +1,63 @@
...
+    const { sourcePath, rootSourcePath, flatModuleFileName } = ngPkg;
+    const rootPathFromSelf = path.relative(sourcePath, rootSourcePath);
 
     return {
       main: this._unixPathJoin(rootPathFromSelf, 'bundles', this._makeUmdPackageName(ngPkg)),
-      module: this._unixPathJoin(rootPathFromSelf, `${ngPkg.fullPackageName}.es5.js`),
-      es2015: this._unixPathJoin(rootPathFromSelf, `${ngPkg.fullPackageName}.js`),
-      typings: path.ensureUnixPath(`${ngPkg.flatModuleFileName}.d.ts`),
-      metadata: path.ensureUnixPath(`${ngPkg.flatModuleFileName}.metadata.json`)
+      module: this._unixPathJoin(rootPathFromSelf, 'esm5', this._makeEsmPackageName(ngPkg)),
```
DavidParks8: We might want to turn `esm5` and `esm2015` into constants so we never have a typo in just one of the places where these folders are used.

src/lib/bundler.ts
```diff
@@ -82,7 +82,7 @@ export async function generateNgBundle(ngPkg: NgPackageData): Promise<void> {
 
   // 9. WRITE PACKAGE.JSON and OTHER DOC FILES
   log.info('Writing package metadata');
-  const packageJsonArtifactPaths: NgArtifacts = artifactFactory.calculateArtifactPathsForPackageJson(ngPkg);
+  const packageJsonArtifactPaths = artifactFactory.calculateArtifactPathsForPackageJson(ngPkg);
```
DavidParks8: Part of the reason for including the explicit typing was so that people could see exactly what was returned from the function without needing an IDE or navigating to the file with the called function. Personally, I find it harder to know what is happening in the code just by reading on github without those explicit typings.

src/lib/steps/uglify.ts
alan-agius4: Do you have any article about this? I'd like to know more :)

PR #326, opened 2017-11-27
perf: copy both `README.md` and `LICENSE` asynchronously togather

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description
combine the async calls together so both IO operations happen at the same time.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ng-packagr.ts
+++ b/src/lib/ng-packagr.ts
@@ -27,14 +27,14 @@ export async function createNgPackage(opts: NgPackagrCliArguments): Promise<void
   let buildDirectoryRoot: string;
   try {
     // READ `ng-package.json` and obtain model, as well as secondary packages
-    const ngPackages: PackageSearchResult = await discoverPackages(opts.project);
+    const ngPackages = await discoverPackages(opts.project);
 
-    const rootPackage: NgPackageData = ngPackages.rootPackage;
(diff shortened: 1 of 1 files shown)


src/lib/ng-packagr.ts
```diff
@@ -27,23 +27,25 @@ export async function createNgPackage(opts: NgPackagrCliArguments): Promise<void
   let buildDirectoryRoot: string;
   try {
     // READ `ng-package.json` and obtain model, as well as secondary packages
-    const ngPackages: PackageSearchResult = await discoverPackages(opts.project);
+    const ngPackages = await discoverPackages(opts.project);
 
-    const rootPackage: NgPackageData = ngPackages.rootPackage;
+    const rootPackage = ngPackages.rootPackage;
```
alan-agius4: I am not sure if these are intentional or not aka, your coding style.
But typescript doesn't need these.

DavidParks8:
This might actually result in a performance degradation depending upon disk type. For spinning disk drives, this can thrash the disk by making the needle jump around a lot. That is why they were not parallel already.

alan-agius4:
Wasn’t aware of that.

Issue #263, opened 2017-11-08
Inline images from scss


mitsoe:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description

When using background-image pointing to an url in a component that is exported the scss is not inlined. This results in not being able to find the image when including the component in a different project.

#### Expected Behaviour

The images are inlined (or a different way to show them) in a consumer module.

#### Version Information

```
ng-packagr: v1.4.1
node: v8.x.y
@angular: v4.2.4
rxjs:5.4.2
zone.js:0.8.14
```

dherges:
A pull request for that would be welcome!

It should support a syntax so that I - as a developer - can decide whether to embed the resource with a `data:` or whether to retain the original url.

PR #345, opened 2017-12-01
feat: inline assets in css file using `data:`

alan-agius4:
Closes: #263

## I'm submitting a...

```
[ ] Bug Fix
[X] Feature
```

## Checklist

## Description
Inline assets such as fonts and images inside the css file.

```
[ ] Yes
[X] No
```


DavidParks8:
Readme documentation for the new field should probably be added.

alan-agius4:
For sure, is there a readme which contains all the options that can be passed?

DavidParks8:
For now, the best we’ve got is the main readme.

README.md
```diff
@@ -168,6 +169,20 @@ For example, the following would use `index.ts` as the secondary entry point:
...
 
+#### How to embed assets in CSS?
+
+You can embed assets such as font and images inside the outputted css. More information [in the CSS tricks website](https://css-tricks.com/data-uris)
+
+```json
+{
+  "ngPackage": {
+    "lib": {
+      "embedAssets": true
```
dherges: Added new "config"

integration/samples/embed-assets/ng-package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "../../../src/ng-package.schema.json",
+  "lib": {
+    "entryFile": "public_api.ts",
+    "embedAssets": true,
```
dherges: Integraton test added.

integration/samples/embed-assets/specs/bundle.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+
+describe(`@sample/inline-assets`, () => {
+
+  describe(`bundle inline-assets.umd.js`, () => {
+    let BUNDLE;
+    before(() => {
+      BUNDLE = fs.readFileSync(
+        path.resolve(__dirname, '..', 'dist', 'bundles', 'inline-assets.umd.js'), 'utf-8');
```
dherges: `require()` the UMD bundle and/or the AoT metadata is better.

integration/samples/embed-assets/specs/bundle.ts
```diff
@@ -0,0 +1,23 @@
...
+    });
+
+    it(`should exist`, () => {
+      expect(BUNDLE).to.be.ok;
+    });
+
+    it(`should have the image inlined with 'data:image'`, () => {
+      expect(BUNDLE).to.contain("data:image/png;base64");
```
dherges: See above. Replace with `METADATA['SomeComponent']['blablubb']['decorators'][0]['styles']`

src/lib/ng-packagr.ts
```diff
@@ -43,6 +43,7 @@ export async function createNgPackage(opts: NgPackagrCliArguments): Promise<void
     }
 
     await copyFiles(`${rootPackage.sourcePath}/README.md`, rootPackage.destinationPath);
+    await copyFiles(`${rootPackage.sourcePath}/CHANGELOG.md`, rootPackage.destinationPath);
```
dherges: New "feat: copies `CHANGELOG.md` to package destination"

src/ng-package.schema.json
```diff
@@ -47,6 +47,11 @@
           "type": "string",
           "enum": ["preserve", "react", "react-native"],
           "default": ""
+        },
+        "embedAssets": {
```
dherges: Schema added! 👍

src/lib/ng-packagr.ts
alan-agius4: Absolutely right, i committed this by mistake :)

README.md
alan-agius4: Can you please explain what you mean please?

alan-agius4:
The postcss-url copy is also a cool option to add tbh. NG CLI and Webpack shouldn't encounter any issue with that.

alan-agius4:
I thought about the ‘copy’ a bit more and actually now I see some problems, the main reason being that the CSS is the JS file, and thus webpack sass-loaders, css-loader won’t be able to parse the css and resolve the url of the image at application level.

I think as in general for library components, I’d rather try to avoid images, and if I really need them i’d go with embed images in order to avoid issues at consumer level.

DavidParks8:
That’s what I ended up doing when I needed an image. It was embedded as an svg in its own component.

alan-agius4:
I’d go as far as having ‘embedAssets’ option set to ‘true’ by default,
considering even that NG CLI is doing so. At least the last time I checked.

My only concern, that might be considered as a Breaking change, but since
there is a V2 coming up maybe it wouldn’t be such a problem.

alan-agius4:
@dherges I am having some problems here with Travis, for some reason the new test is failing, however it does work locally and on circleci

dherges:
It says `Error: Cannot find module '../dist/embed-assets.metadata.json'`

Can you rebase the changes on master and update the PR?

alan-agius4:
I already did that, and seemed to work locally and on circle. but i can
give it another shot tomorrow.

alan-agius4:
All green now :)

alan-agius4:
Should we set this to true by default?

dherges:
Hm. Maybe. First, let's change the property to an enum:

```
cssUrl: 'inline' | 'none'
```

alan-agius4:
@dherges property changed

PR #356, opened 2017-12-03
feat: implement `transformSources()` w/ domain model

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Introduces a domain model for _Package_ and _Entry Point_. Serves as a foundation for a set of interesting ideas: what if... we could analyze an entry points `import` statements and dynamically create a minimal rollup configuration? What if... we further can detect what entry points depends on another entry point (within the same package) and we could create a build graph and automatically adjust the tsconfig?

BREAKING CHANGES:
Introduces a domain model for _Package_ and _Entry Point_ (as defined in Angular Package Format Specification Glossary). Albeit the refactoring was undertaken with care and integration tests were not changed, it may cause undesired behaviour.

## Notes

Technically, this is not a feature but a refactoring. Marking it as "feat:" with "BREAKING CHANGES" for auto changelog reasons.

```
[x] Yes
[ ] No
```


dherges:
TODO:
- [ ] add legacy `esmBundleFile`, `umdBundleFile` property to `NgEntryPoint`, straight away `@deprecated`
- [ ] let tests turn green (to verify no regression)
- [ ] open a new PR for "consistent flatModuleFileName" instead of legacy bundle names

Issue #371, opened 2017-12-06
TypeScript helpers are inlined in FESM5 and FESM2015 package


alan-agius4:
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description
Following the new RC and this PR #338. Ng Packagr now supports typescript helpers. however these are are being inlined inside the FES5 modules, and not external like https://github.com/angular/common-builds/blob/master/esm5/common.js#L7, 

Also, I noticed that tslib, is not marked as `peerDependency` nor `devDependency` in your package.json. Thus, ng-packagr is working because any `Angular` package has `tslib` as dependency.
https://github.com/angular/common-builds/blob/master/package.json#L12

#### How To Reproduce
Run `npm t` in ng-packagr or build a normal library

#### Expected Behaviour
tslib is only inlined for UMD bundles
tslib is added as a dependency in `package.json`
tslib is set as an external in FESM5

#### Version Information
```
ng-packagr: 2.0.0-rc.5 
```

_please include any version information that might be relevant, e.g. other third-party libraries_

unlight:
Why tslib are embeded in UMD bundle?
Is it possible to make tslib external for UMD also?

alan-agius4:
That’s following the APF specs.

One of the reasons is that users shouldn’t care if a library is built with
TypeScript or not.

PR #372, opened 2017-12-06
fix: `TypeScript` helpers are inlined in FESM5 and FESM2015 package

alan-agius4:
Closes: #371

## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

- tslib contains should only be inlined for `UMD` bundles while still referencing the external library in ES5 and ES2015.
- tslib added as a dependency in the final `package.json`

```
[ ] Yes
[x] No
```


dherges:
In general, lgtm!

Note to self: we need unit tests to write better specs for the individual build steps. And a common `interface BuildStep {  (...buildArgs): Promise<...> }` ...

dherges:
Can we add a test that verifies `tslib` is imported in the FESMs and inlined in the UMDs?

alan-agius4:
sure, pushed some more tests.

alan-agius4:
Need to check a bit as with `tsickle` this broke

src/lib/entry-point-transforms.ts
```diff
@@ -59,20 +64,32 @@ export const transformSources =
...
+      }
     });
-    await remapSourceMap(fesm15File);
 
     // 4. FESM5: TSC
     log.info('Bundling to FESM5');
     const fesm5File = path.resolve(artefacts.stageDir, 'esm5', entryPoint.flatModuleFile + '.js');
     await downlevelWithTsc(fesm15File, fesm5File);
-    await remapSourceMap(fesm5File);
+    await rollup({
```
alan-agius4: this is required because without rollup helpers won't be cherry picked.

ie: it will be 
```js
import * as tslib from 'tslib';
```
instead of
 
```js
import { __extends } from 'tslib';
```

src/lib/entry-point-transforms.ts
```diff
@@ -84,7 +101,6 @@ export const transformSources =
       dest: umdFile,
       externals: entryPoint.externals
     });
-    await remapSourceMap(umdFile);
```
alan-agius4: we don't need to remapSourcemaps after `rollup` we can do that with rollup

alan-agius4: Also, without this, after i run rollup on the downleveled sorcery would fail

src/lib/entry-point-transforms.ts
```diff
@@ -59,19 +64,35 @@ export const transformSources =
...
+      }
     });
     await remapSourceMap(fesm15File);
 
     // 4. FESM5: TSC
     log.info('Bundling to FESM5');
     const fesm5File = path.resolve(artefacts.stageDir, 'esm5', entryPoint.flatModuleFile + '.js');
-    await downlevelWithTsc(fesm15File, fesm5File);
+    const downLevelTmp = path.resolve(pkg.workingDirectory, entryPoint.flatModuleFile + '.js');
+    await downlevelWithTsc(fesm15File, downLevelTmp);
```
alan-agius4: this is required because without rollup typescript helpers are not cherry picked.

ie it will be;
```
import * as tslib1 from 'tslib';
```

rather than
```ts
import { __extends } from 'tslib';
```

src/lib/entry-point-transforms.ts
```diff
@@ -59,19 +64,35 @@ export const transformSources =
...
+        [TS_HELPERS_LIB.name]: TS_HELPERS_LIB.name
+      }
     });
     await remapSourceMap(fesm15File);
 
     // 4. FESM5: TSC
     log.info('Bundling to FESM5');
     const fesm5File = path.resolve(artefacts.stageDir, 'esm5', entryPoint.flatModuleFile + '.js');
-    await downlevelWithTsc(fesm15File, fesm5File);
+    const downLevelTmp = path.resolve(pkg.workingDirectory, entryPoint.flatModuleFile + '.js');
```
alan-agius4: Not sure if this is the right place to have a tmp file.

dherges:
In response to: https://github.com/dherges/ng-packagr/pull/372#issuecomment-350441972 ...

I wouldn't focus too much on the `tslib` helpers right now. It's an optimization and "we get await with it" w/o the optimization. There are other issues / PRs who need more attention :-) especially rollup, then improving test cases, the `es*` language level and I have some fixes and ideas I'd like / need to bring in...

alan-agius4:
Ill pause this for be the time being at least until the rollup PR is merged, as it will most likely effect this, and with that PR it will make this a bit simpler as well.

alan-agius4:
Closed this in flavor of #428

Issue #190, opened 2017-10-18
[PR WELCOME] Discover primary and secondary entry points from user configuration


dherges:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Historically, a `package.json` and a `ng-package.json` are used to "configure" ng-packagr. The `package.json` is used for the module ID, npm metadata, and more node/npm-related things. The `ng-package.json` is used for destination paths, finding the entry file of an entry point, and angular/typescript/rollup-specific things.

Entry point discovery should be changed, see below. Discovery of primary and secondary entry points should both use this.

#### Expected Behaviour

New Draft / Outline (first one "wins"):

 - `package.json`: `ngPackage`
 - `ng-package.json` (requires a `package.json` as sibling)
 - `ng-package.js` (default export, requires a `package.json` as sibling)

Note: ignore `package.json` w/o `ngPackage` property.

#### Version Information

```
ng-packagr: v2.0.0-rc.0
```

DavidParks8:
I like this idea. If we implement it, then we should probably release a 1.x version which has a deprecation warning when it's used with ng-package.json.

dherges:
One aspect makes me feel a little doubtful: it feels like polluting everything with `pom.xml` - just the name is now `package.json`. Technically, for npm, it would encourage to produce a lot of "invalid" package.json files (files w/o name, version, and so on...).

Haven't tried ... what happens when `cd path/to/some/package-json && npm install`?

DavidParks8:
Also, I didn't document it, but `ng-package.json` should work as a marker for the creation of a secondary package as well. I don't know if that changes anything. Time for work! I'll check the npm sub install thing later.

dherges:
For the primary entries in a "monorepo project" with multiple packages, I added `"scripts": { "preinstall": "exit 1" } in each `package.json` to prevent `npm install` / `yarn install` in the subfolders. So I don't know what happens with empty `package.json` files at all.

I am also interested how the npm teams sees the matter of putting multiple `package.json` files in one project ... have not any info for the moment though. Is there some blog post, issue discussion or similar docs on it?!?

DavidParks8:
I just found https://stackoverflow.com/questions/10065564/add-custom-metadata-or-config-to-package-json-is-it-valid in support of custom package.json fields, but the more I think about it, the more I want to preserve both configuration methods. People are opinionated and like to choose for themselves how they configure things, and forcing them into one configuration method might make many unhappy.

[...]

PR #383, opened 2017-12-08
feat: discover entry points from user configuration

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Discovery of primary and secondary entry points have been changed to the following;
 - `package.json`: with `ngPackage` property
 - `ng-package.json` (requires a `package.json` as sibling)
 - `ng-package.js` (default export, requires a `package.json` as sibling)

Closes: #190

```
[ ] Yes
[x] No
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,12 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "sample: package-js",
+      "program": "${workspaceRoot}/integration/samples/package-js/build.dev.js"
+    },
+    {
+      "type": "node",
+      "request": "launch",
       "name": "sample: same-name",
       "program": "${workspaceRoot}/integration/samples/same-name/build.dev.js"
     },
--- a/README.md
+++ b/README.md
@@ -85,9 +85,25 @@ What about [ng-packagr alongside Nx Workspace](https://github.com/dherges/nx-pac
 
 #### Configuration Locations
 
-Configuration is picked up from the cli `-p` parameter, then from `package.json`, then from `ng-package.json` and lastly from `ng-package.js`.
+Configuration is picked up from the project file given by the `-p` CLI option.
+The `-p `option may refer to a `package.json` (with custom `ngPackage` property), an `ng-package.json`, or an `ng-package.js` file.
+When the `-p` option refers to a directory, the configuration is picked up from the first matching source;
+locations are tried in the above-mentioned order.
 
-To configure with a `ng-package.json` or `ng-package.js`, put the `package.json` of the library in the same folder next to the `ng-package.json` or `ng-package.js`.
+To configure with a `package.json`, put the configuration in the `ngPackage` custom property:
+
+```json
+{
+  "$schema": "./node_modules/ng-packagr/package.schema.json",
+  "ngPackage": {
+    "lib": {
+      "entryFile": "public_api.ts"
+    }
+  }
+}
+```
+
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -85,10 +85,11 @@ What about [ng-packagr alongside Nx Workspace](https://github.com/dherges/nx-pac
 
 #### Configuration Locations
 
-Configuration is picked up from the cli `-p` parameter, then from the default location for  `ng-package.json`, then from `package.json`.
+Configuration is picked up from the cli `-p` parameter, then from `package.json`, then from `ng-package.json` and lastly from `ng-package.js`.
```
dherges: A little misleading.

> Configuration is picked up from the project file given by the cli `-p` option.
> The `-p` option may refer to a `package.json` (with custom `ngPackage` property), an `ng-package.json`, or an `ng-package.js` file. 
> When the `-p` option refers to a directory, the configuration is picked up from the first matching source; the locations are tried in the above-mentioned order.

.vscode/launch.json
```diff
@@ -13,6 +13,12 @@
       "name": "sample: material",
       "program": "${workspaceRoot}/integration/samples/material/build.dev.js"
     },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "sample: package-js",
+      "program": "${workspaceRoot}/integration/samples/package-js/build.dev.js"
```
dherges: Is ok. I'll change it to the general-purpose script ;-)

integration/samples/package-js/secondary/ui-lib.module.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@NgModule({
+  declarations: [
+    BazComponent,
+  ],
+  exports: [
+    BazComponent,
+  ]
+})
+export class UiLibModule {}
```
dherges: Think we should keep the sample really minimalistic as it's just there to verify the conf location... is it ok I'll reduce it after merging?

src/lib/steps/init.ts
```diff
@@ -33,73 +34,48 @@ interface UserPackage {
...
+      packageConf = undefined;
+    }
 
-        // TODO: enable this, see #278
-        if (4 == 4) {
-          throw new Error(`Reading 'ngPackage' from a .js file is not yet implemented.`);
-        }
+    if (packageConf || pathStats.isDirectory()) {
+      // return even if it's undefined and use defaults when it's not a file
+      return packageConf;
```
dherges: I don't understand this. `packageConf` is initailized, then set to `undefined` in the "else", then checked here for "truthy" ... but ok as long as it works!?! Oo :)

alan-agius4: Tried to keep it simly, maybe i can do it in reverse if you’d prefer. Ie start as undefined and if not undefined, i’ll assign properties to the object.

integration/samples/package-js/secondary/ui-lib.module.ts
alan-agius4: Yep absolutely

src/lib/steps/init.ts
alan-agius4: Ill see what I can do tomorrow or maybe merge it and I’ll refactor it :)

dherges: No, it's ok ... yes, maybe it would be better to initialize with `undefined` and then assign values if it finds something... matter of taste. I'll merge it for now, ok?

alan-agius4:
Sounds good to me, I’ll try to find some time tomorrow to refactor this a
bit :)

Issue #312, opened 2017-11-23
[PR WELCOME] Dynamic rollup configuration // print descriptive error messages


DavidParks8:
## Type of Issue

```
[x] Feature Request
```

## Description

Many people seem to have the same issue with various libraries: they don't have correct rollup externals.

#### How To Reproduce

Look at #310, #301, #85, #119, #108, #142, #230, #163, etc... 

#### Expected Behaviour

I suspect that a lot of these issues can be avoided by printing a more helpful error message.

#### Version Information

```
ng-packagr: v2.x.y
node: v8.x.y
@angular: v4.x.y
rxjs:
zone.js:
```

alan-agius4:
Question, why not have externals in rollup be a function like the below so that all node modules are treated  as externals 

‘’’
externals: (path) => path.indexOf(“node_modules”) !== -1;
‘’’

matheo:
I got a "descriptive" error message but I wonder what can I do about it.
```
$ node_modules/.bin/ng-packagr
Building Angular library
Generating bundle for @coachcare/common
Cleaning bundle build directory
Processing assets
Running ngc
Compiling to FESM15
'http' is imported by node_modules/axios/lib/adapters/http.js, but could not be resolved – treating it as an external dependency
'https' is imported by node_modules/axios/lib/adapters/http.js, but could not be resolved – treating it as an external dependency
preferring built-in module 'url' over local alternative at '/cli-project/node_modules/url/url.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
preferring built-in module 'url' over local alternative at '/cli-project/node_modules/url/url.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
'url' is imported by node_modules/follow-redirects/index.js, but could not be resolved – treating it as an external dependency
preferring built-in module 'assert' over local alternative at '/cli-project/node_modules/assert/assert.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
preferring built-in module 'assert' over local alternative at '/cli-project/node_modules/assert/assert.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
'assert' is imported by node_modules/follow-redirects/index.js, but could not be resolved – treating it as an external dependency
'http' is imported by node_modules/follow-redirects/index.js, but could not be resolved – treating it as an external dependency
[...]

PR #395, opened 2017-12-11
feat: dynamic rollup configuration for esm flattening

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[X] Feature
```

## Checklist

## Description
`externals` has been removed in favor of `namedExternals` which is now only used to name the external libraries.

By default all dependencies are now treated as externals and thus are not embedded in the final bundle.

However, If you want to embed one or more of them into the final bundle you can do so by adding the library in the `embeeded` section like so;

```json
{
  "$schema": "../../../src/ng-package.schema.json",
  "lib": {
    "embedded": [
      "lodash",
      "date-fns"
    ]
  }
}
```

While `rollup` will do it's best to guess the name of an external library some modules are hard to guess and it's recommended to provide the name of the external dependency. This can be done by using `namedExternals` in the config section like so;

```json
{
  "$schema": "../../../src/ng-package.schema.json",
  "lib": {
    "namedExternals": {
      "lodash" : "_",
      "date-fns" : "DateFns",
    }
  }
}
```

```
[X] Yes
[ ] No
```

Closes: #312


src/lib/steps/rollup.ts
```diff
@@ -23,37 +23,31 @@ export interface RollupOptions {
...
   log.debug(`rollup (v${ROLLUP_VERSION}) ${opts.entry} to ${opts.dest} (${opts.format})`);
-
-  const globals = {
-    // default externals for '@angular/*' and 'rxjs'
-    ...ROLLUP_GLOBALS,
-    // external symbols passed from the user's ng-package.json
-    ...opts.externals,
+  const embedded: string[] = opts.embedded || [];
+  const namedExternals = {
+    'tslib': 'tslib',
```
alan-agius4: might also make sense to add some other defaults that are widely used example, `ionic-angular`, `lodash`, `moment`, `date-fns`, `react` etc..

alan-agius4:
Need to update lockfile, though I am on my mobile at the moment!

src/lib/steps/rollup.ts
```diff
@@ -22,38 +21,32 @@ export interface RollupOptions {
...
-        }
+      if (
+        path.isAbsolute(moduleId) ||
+        moduleId.startsWith(".") ||
+        moduleId.startsWith("/") ||
+        moduleId.indexOf("commonjsHelpers") >= 0 || // in case we are embedding a commonjs module we need to include it's helpers also
+        embedded.some(x => x === moduleId)
+      ) {
+        // if it's either 'absolute', marked to embed, starts with a '.' or '/' it's not external
+        return false;
```
dherges: What do you think of extracting to a function / lambda?

```ts
const externalModuleIdStrategy = (moduleId: string): boolean => {
  /* ... */
}
```

Then write a unit test for it?

alan-agius4: I’d love that, if you tell me where to place the unit test etc... :) I’d find it much more easy to test.

src/lib/steps/rollup.ts
```diff
@@ -77,7 +70,42 @@ export async function rollup(opts: RollupOptions): Promise<void> {
...
+      if (/^rxjs\/symbol/.test(moduleId)) {
+        return "Rx.Symbol";
+      }
+
+      if (/^rxjs\/(add\/)?operator/.test(moduleId)) {
+        return "Rx.Observable.prototype";
+      }
+
+      return namedExternals[moduleId] || ''; // leave it up to rollup to guess the global name
+    },
```
dherges: See above. Extract to a function / named lambda and then write a unit test for that piece of functionality?

```ts
const umdModuleIdStrategy = (moduleId: string): string => { /* ... */ }

/* ... */

describe(`umdModuleIdStrategy`, () => {
  it(`should map rxjs/add/operator/<foo> to Rx.Observable.<foo>`, () => {
    /* ... */
  });
});
```

???

dherges: Wdyt?

src/lib/steps/rollup.ts
alan-agius4: Small correction from your example, to test it, you need to export it though!

dherges:
Can you try to reset to `origin/master`? Ideally it should be enough to keep the changes to these files:

screen shot 2017-12-11 at 21 40 48


Unit tests have been added on the master, command: `yarn test:specs`

The test file is named `<being-tested>.spec.ts`. Keep that pattern with `rollup.ts` and `rollup.spec.ts`.

dherges:
Here's an example:

https://github.com/dherges/ng-packagr/blob/0079deb38f64320ab71d02d3f4c0056c5bf08ca9/src/lib/util/ts-transformers.spec.ts#L1-L9

src/lib/steps/rollup.ts
dherges: Ah, yes! 😄

alan-agius4:
Thanks, will do that.

alan-agius4:
@dherges Unit tests added

alan-agius4:
Bump!

dherges:
From [rollup's API docs](https://rollupjs.org/#big-list-of-options):

> globals: Object of id: name pairs, used for umd/iife bundles.

Let's rename `namedExternals` to`umdModuleIds` since it's only intention is for helping rollup to find umd module ids.

src/lib/steps/rollup.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    const FOO_MODULE = 'FooModule';
+    it(`should map 'foo' to '${FOO_MODULE}' when 'namedExternals' is provided`, () => {
+      expect(umdModuleIdStrategy('foo', { 'foo': FOO_MODULE})).to.equal(FOO_MODULE);
+    });
+
+    it(`should map 'foo' to '' when no 'namedExternals' is provided`, () => {
+      expect(umdModuleIdStrategy('foo')).to.empty;
+    });
+  });
+});
```
dherges: 👍

integration/samples/rollup/README.md
```diff
@@ -0,0 +1,2 @@
+Sample library: external and embeded dependencies 
+=====================================
```
dherges: Ok, I think we can move this integration test to `integration/samples/external`

dherges:
@alan-agius4 When you make these two little adjustments, I think it's good to go!

alan-agius4:
@dherges updated as requested.

dherges:
@alan-agius4 perfect!

dherges:
Just added a few words to the readme, hoping to clarify and be precise on the terms "externals", "dependency", "bundle", and so on

dherges:
Thank you Alan, this is a good step to the right direction!

alan-agius4:
Glad to help :)

alan-agius4:
@dherges any other ideas that I can help with? (Sry to write here in the PR but I cant find your email)

Issue #379, opened 2017-12-07
[DRAFT / DISCUSSION] Intra-package dependencies (between entry points)


dherges:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

> I wonder how to handle the dependencies here. I have a `@org/components` root package, and I would like to have `@org/components/shared` for shared utilities and stuff.
> The root package can import stuff from shared without generating duplicated compile code?
> Is there a way to set up this kind of "dependency"?

## Motivation

Allow dependencies between entry points of the same package.

## Expected Behaviour

By TypeScript AST Traversal, detect whether `@org/components` depends on another entry point within the same package, e.g. `@org/components/shared`.

Then create a build graph, i.e. `@org/components/shared` is required to be built before `@org/components`.

Then, we run the transformation process for packages w/o inter-entry-point deps first. When running the transformation process for dependent packages, configure tsc `compilerOptions.paths` to point to a FESM bundle of its dependency. 

## Version Information

```
ng-packagr: 2.0.0-rc.x
```

dherges:
What happens if we detect a cyclic dependency?

My view: error and non-zero exit code.

DanielSchaffer:
I realize this discussion was just started 6 hours ago, but any ideas as to when this would be released? I could absolutely use it now! Also, sorry for the meta-question, is there a better place to ask something like this?

dherges:
No, no idea. No, there is no good place for such questions.

Continue on the topic!

Cheers, David

alan-agius4:
From my point of view, if you have have something that can be shared/re-used I'd say create a separate package rather than a separate entry point. But that's just my opinion.

And yeah with `compilerOptions.paths` it is possible to be done, I am doing something similar for a mono-repo

dherges:
So I started a first draft in #419. Ideas and thoughts?

PR #419, opened 2017-12-17
[WIP] feat: intra-package dependencies

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Draft for #379 

```
[ ] Yes
[x] No
```


dherges:
Open / Todos:
- rollup warnings since `external` is unknown (for entry point dependency)
- UMD bundle errorneous since UMD module ID incorrect (for entry point dependency)
- skipped secondary builds: start from scratch again. re-suming previous build is becoming pretty hard-coded.
- integration tests fail because it can't resolve secondary entry point imports, i.e. `import {..} from '@sample/secondary-lib/shared-module` cannot be resolved

DanielSchaffer:
Not sure if this is a separate issue or not, but I noticed while trying to build with this branch that it seems to have two regressions:
- it would not read the `ngPackage` config for the main entry point from `package.json` - it had to be in `ng-package.json` (error was `ENOENT: no such file or directory, lstat '/Users/daniel/Workspace/siplay-ng-system/_submodules/ng-common/ng-package.json'`)
- Similarly, while the presence of a `package.json` with the `ngPackage` property in a subdirectory is correctly interpreted as a secondary entry point, it seems to ignore the `lib` configuration there, using the defaults for the entry file (apparently `src/public_api.ts`?) Putting the config in an `ng-package.json` file works. Oddly enough, skipping the secondary entry point's `package.json` file entirely in favor of `ng-package.json` doesn't work either, resulting in the warning `Cannot read secondary entry point at /Users/daniel/Workspace/siplay-ng-system/_submodules/ng-common/config. Skipping.`

I'm not sure whether dependency builds are actually supposed to be working at this point of the WIP, but I'm still getting `Could not resolve entry` errors on any intra-package dependency.

matheo:
@dherges I've solved the internal dependencies with a manual `NgPackage.paths` field, filling it manually with the relative paths (from the intra-package to the other intra-modules) and the same with `lib.externals`. 

I think we can autogenerate them detecting the intra-modules and have a convention about the folder structure to avoid cycles. Like autogenerate the paths and externals for sibling folders and not parent folders. There may be edge cases but with this convention, we can prevent some bad practices. Let's take an example:

```
common (@org/common)
 |- api (@org/common/api)
 |- components (@org/common/components)
 |- shared (@org/common/shared)
    |- config (@org/common/shared/config)
```

Adding the paths and externals on the NgPackage object for `@org/common/api`:
```
{
    ...
    "paths": {
      "@org/common/components": ["../components/index.ts"],
      "@org/common/shared": ["../shared/index.ts"],
      "@org/common/shared/config": ["../shared/config/index.ts"]
    },
    "lib": {
      "entryFile": "index.ts",
      "externals": {
        ...
        "@org/common/components":"@org/common/components",
        "@org/common/shared": "@org/common/shared",
        "@org/common/shared/config": "@org/common/shared/config"
      }
    }
  }
}
```

And for `@org/common/shared/config` `ng-packagr` should add:
```
{
    ...
    "paths": {
      "@org/common/api": ["../api/index.ts"],
      "@org/common/components": ["../components/index.ts"]
    },
    "lib": {
      "entryFile": "index.ts",
      "externals": {
        ...
        "@org/common/api":"@org/common/api",
        "@org/common/components":"@org/common/components"
      }
    }
  }
}
```
Note that it will avoid its branch, as theoretically it should not depend on its parents.
If there's a `Could not resolve entry` the dependencies are not organized as suggested.
Just my 0.02 ;)

alan-agius4:
For me personally I am facing the same issue however, i am working in a mono repo structure. For me personally, all i’d would want is to have access to provide ‘paths’

dherges:
Please keep the general discussion in the issue.

src/lib/domain/ng-package-format.ts
```diff
@@ -163,6 +168,15 @@ export class NgEntryPoint {
...
   }
 
+  /**
+   * TODO
+   * Dependencies of this entry point on other entry points in the package.
+   * List of module IDs
+   */
+  public dependendingOnEntryPoints: string[] = [];
+
+  public buildStatus: 'failed'|'success'|'pending'|'inprogress' = 'pending';
```
alan-agius4: My i suggest to use string enums rather than types?

enum BuildStatus {
    InProgress = ‘InProgress’
}

Usage would be ‘BuildStatus.InProgress’ 

Unless you want to use normal enums.

Sry can only write messages, as I am in a hurry!

src/lib/entry-point-transforms.ts
```diff
@@ -36,9 +37,34 @@ export const transformSources: BuildStep =
...
-    collectTemplateAndStylesheetFiles(args);
+    // First pass: extract templateUrl and styleUrls, analyse dependencies
+    analyseSourceFiles(args);
+
+    // XX: here starts the real work :-)
+    if (entryPoint.dependendingOnEntryPoints.length > 0) {
+      log.info(`${entryPoint.moduleId} depends on ${entryPoint.dependendingOnEntryPoints.join(', ')}.`);
+
+      // Were the depending entry points already built?
+      const unsatisfiedDependencies = entryPoint.dependendingOnEntryPoints
```
DavidParks8: I might be missing something here, but wouldn't this be easier if you construct the build graph, determine the order that things should be built, then build them in that order without checking whether or not dependencies have already been built? I think doing so would remove the need to code up build resuming of specific entry points.

DavidParks8:
Let me first preface this by saying that the code below is an untested first draft, and I will take no offense if you don't use it. It might be helpful, it may also be buggy. If you would like, I'm willing to make a PR that adapts it to the current code and ensures it works with tests. To be honest, making a dependency graph was a fun problem, so I felt like solving it. Here is a set of classes that can be used to create a graph of dependencies, and build them in parallel (when possible) without the need to resume previously attempted builds. 

```ts
export class NgProjectToken {
  public path: string;
  public name: string;
}

/**
 * The projects within a parallel build group can be safely built in parallel.
 */
export class ParallelBuildGroup extends Array<NgProjectToken> {
}

/**
 * Serves as an internal representation of a node in the build graph.
 */
class BuildNode {
  private readonly _parents: Set<BuildNode> = new Set<BuildNode>();
  public get parents(): Set<BuildNode> {
    return this._parents;
  }

  private readonly _project: NgProjectToken;
  public get project(): NgProjectToken {
    return this._project;
  }

  constructor(project: NgProjectToken) {
    this._project = project;
  }
}

export class ProjectGraphBuilder {
  private readonly _nodes: Map<NgProjectToken, BuildNode> = new Map<NgProjectToken, BuildNode>();

  public addProject(project: NgProjectToken, dependencies?: NgProjectToken[]): void {
    const projectNode: BuildNode = this._getOrCreateNode(project);

    if (dependencies) {
      for(const dependency of dependencies) {
        const dependencyNode: BuildNode = this._getOrCreateNode(project);
        if (dependencyNode.parents.has(projectNode)) {
          // todo: detect larger than nearest-neighbor circular loops
          throw new Error(`Circular dependency detected between ${projectNode.project.name} and ${dependencyNode.project.name}.`);
        }


        projectNode.parents.forEach((projectParent: BuildNode) => {
          if (dependencyNode.parents.has(projectParent)) {
            // erase local reference, because the parent already has one
            dependencyNode.parents.delete(projectParent);
          }
        });

        projectNode.parents.add(dependencyNode);
      }
    }    
  }

  public build(): ParallelBuildGroup[] {
    // this serves as a tracker for what nodes have been processed, and their furthest distance from a root node
    const nodeDepths = new Map<BuildNode, number>();
    const allNodes: BuildNode[] = Array.from(this._nodes.values());
    const nodeQueue: BuildNode[] = allNodes.filter(x => x.parents.size === 0); // start with root nodes and expand from there

    for(const node of nodeQueue) {
      nodeDepths.set(node, 0);
    }

    // We will use this later to reduce the frequency of dynamic memory allocations
    let deepestDepth: number = 0;

    // Walk the tree to determine longest path from root for each node
    while(nodeQueue.length !== 0) {
      const currentNode: BuildNode = nodeQueue.pop();
      const currentDepth: number = nodeDepths.get(currentNode) + 1;
      if (currentDepth > deepestDepth) {
        deepestDepth = currentDepth;
      }

      currentNode.parents.forEach((dependency: BuildNode) => {
        if (!nodeDepths.has(dependency)) {
          nodeQueue.push(dependency);
          nodeDepths.set(dependency, currentDepth);
        }

        const dependencyDepth: number = nodeDepths.get(dependency);
        if (currentDepth > dependencyDepth) {
          nodeDepths.set(dependency, currentDepth);
        }
      });
    }

    // If we wanted to possibly speed up the build, we could look at all root nodes and build a flat list of their dependencies.
    // If the intersection of dependencies between two roots is empty, then the roots can be built independently.
    // For now, this speed up seems unnecessary, so it is omitted for the sake of reducing complexity.

    // All nodes with the same max distance from a root can be run in parallel
    // Now we need to bucket nodes by max depth
    return this._bucketIntoParallelGroups(nodeDepths, deepestDepth);
  }

  private _bucketIntoParallelGroups(nodeDepths: Map<BuildNode, number>, maxDepth: number): ParallelBuildGroup[] {
    maxDepth++;
    const buckets: ParallelBuildGroup[] = new Array<ParallelBuildGroup>(maxDepth);
    for(let i = 0; i < maxDepth; i++) {
      buckets[i] = new ParallelBuildGroup();
    }

    nodeDepths.forEach((depth: number, node: BuildNode) => {
      buckets[depth].push(node.project);
    });
    
    return buckets;
  }

  private _getOrCreateNode(project: NgProjectToken): BuildNode {
    if (this._nodes.has(project)) {
      return this._nodes.get(project);
    }

    const newNode: BuildNode = new BuildNode(project);
    this._nodes.set(project, newNode);
    return newNode;
  }
}

// An example of how this could be used:
// export async function buildIt() {
//   const builder = new ProjectGraphBuilder();
//   // add dependencies to graph here...

//   const groups = builder.build();
//   for(const independentGroupOfProjects of groups) {
//     await Promise.all(independentGroupOfProjects.map((project: NgProjectToken) => {
//       // run the entire ng-packagr build pipeline here, since each project needs to be independently packaged
//       return Promise.resolve();
//     }));
//   }
// }
```

jolafrite:
Hi guys, how would you manage this with the current version of ng-packagr ?

scttcper:
Here's what I'm doing to build modules with two common packages https://github.com/scttcper/ngx-color/blob/master/build.ts

While I think it would be cool if it can be detected, I think passing an array of packages to build in order would be fine.

dherges:
Hi all,

@scttcper's example is a pretty good demo how you can do it at the moment. Let me say that it still feels like a workaround and a proper solution is likely to come not before 2.0.0

dherges:
@DavidParks8 your ideas of nodes and graphs have made it to the re-write ;-)

You're saying to build the graph and analyse dependencies between nodes before performing the build. When you had that in mind, that means to do the TypeScript AST traversal on each entry point first and only after that continue with subsequent processing (rendering stylesheets, compiling ngc, and so on...). Is it that what you're saying?

DavidParks8:
@dherges that is exactly what I had in mind :)

dherges:
@DavidParks8 Hm. THat's a two side sword.

Mapping paths to sources:
- downside: it's building against sources, compile errors of dependent may show in depending
- upside: one typescript programm in memory at a time

Mappings paths to dists:
- upside: building against sources
- downside: multiple typescript programs in memory, analysing sources up front

At least initializing the typescript programs is quite expensive for I/O (guess since reading many small files). The question is what impact keeping several typescript compilers in memory has ...

matheo:
@dherges currently the intra-packages uses the paths of the root package or is this dependency map not solved in 2.1.0?

dherges:
Hi @DavidParks8, in your code from https://github.com/dherges/ng-packagr/pull/419#issuecomment-354015908, what is the depth? Is that the number of "dependency edges"?

Example:
- foo dependsOn bar
- foo dependsOn foobar
- bar dependsOn foobar

foobar has depth 0 (no dependency), bar has depth 1, foo has depth 3 (2 from the dependency on bar, 1 from the dependency on foobar)?

DavidParks8:
image

DavidParks8:
The example I used to design the algorithm was the following set of dependencies: 
image

Where A depends on B which depends on C which depends on D etc. This example would yield the following build graph: 
image

dherges:
Ok, think I got it. The depth is then the index in the array. I've got the dependency analysis in place now. I will take your code scribbles and weave it into the new pipeline, replacing [this naive concat(...projects)](https://github.com/dherges/ng-packagr/blob/master/src/lib/ng-v5/package.transform.ts#L61-L66) with the order determined by your code!

Btw, you changed your profile picture! Gives a face to the person! :+1:

DavidParks8:
Sweet! Yeah my job at Microsoft required that I show a real picture in order to join the official github organization.

matheo:
@dherges @DavidParks8 thank you both!!

PR #428, opened 2017-12-20
fix: `TypeScript` helpers are inlined in FESM5 and FESM2015 package

alan-agius4:
## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

- tslib contains should only be inlined for `UMD` bundles while still referencing the external library in ES5
- tslib added as a dependency in the final `package.json`

Closes: #371

```
[ ] Yes
[x] No
```


alan-agius4:
@dherges bump

alan-agius4:
@dherges  bump

alan-agius4:
@dherges bump again!

emoralesb05:
Would this be getting in soon? I would love if this was merged before stable 😄 🎉

stevenov7:
joining the **bump** party! 😄 🎉

alan-agius4:
@dherges can we have an answer about this please? Thanks

alan-agius4:
So how do you think it's best to proceed?

alan-agius4:
@dherges I updated the PR, after your changes as I still noticed some things are missing

dherges:
Hi @alan-agius4,

can you quickly talk me through how the change is supposed to work?

The default tsconfig includes:

```json
  "compilerOptions": {
    "importHelpers": true,
  }
```

After ngc compilation, there are individual es2015 js files and the [`externalModuleIdStrategy`](externalModuleIdStrategy) should treat tslib as external by default. Was this only affecting the es2015 to es5 downlevelling?

src/lib/flatten/tsc.ts
```diff
@@ -17,6 +17,7 @@ export async function downlevelWithTsc(inputFile: string, outputFile: string): P
     target: ScriptTarget.ES5,
     module: ModuleKind.ES2015,
     allowJs: true,
+    importHelpers: true,
```
dherges: ES2015 to ES5 down-levelling.

src/lib/flatten/flatten.ts
```diff
@@ -70,7 +71,7 @@ export async function flattenToUmd(opts: FlattenOpts): Promise<string> {
     umdModuleIds: {
       ...opts.umdModuleIds
     },
-    embedded: opts.embedded
+    embedded: ['tslib', ...embedded]
```
dherges: Add `tslib` to embedded dependencies in UMD flattening.

src/lib/flatten/tsc.ts
alan-agius4: Didn't quite get what you mean, you need to importHelpers while downleveling as without it it won't use `tslib`

alan-agius4:
hey @dherges,

`tslib` should only be treated as external for consumers not using `umd` For umd modules, `tslib` should be embedded, as in reality anyone using a library shouldn't care if it was written in typescript or not.

While for ESM5 users they will probably use a module loader, like webpack for example which is able to resolve `tslib` and share it's methods for all usages and all libraries and thus no need for it to be embedded. 

In a nutshell
- tslib should be embeeded in UMD
- tslib should be external for ESM5 modules
- No tslib for ESM2015
- add tslib as a dependency in `package,json`, this is important as without this the consumer will break unless there is another library that uses `tslib`

I'll try to fix the issues in the PR today or tomorrow

src/lib/flatten/tsc.ts
dherges: Is the change affecting just the two code occurences?

1. https://github.com/dherges/ng-packagr/pull/428#discussion_r165038381
2. https://github.com/dherges/ng-packagr/pull/428#discussion_r165038237

alan-agius4: And https://github.com/dherges/ng-packagr/pull/428/files#diff-a4eee2c0b9136c939507b9942ac400a9R30

https://github.com/dherges/ng-packagr/pull/428/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R56

and added some integration tests.

alan-agius4:
I opened another PR, as not to have the formatting stuff as I noticed that it was due to merging and rebasing.

https://github.com/dherges/ng-packagr/pull/573

Issue #425, opened 2017-12-19
[Bug] Unicode in scss property


appienvanveen:
## Type of Issue

```
[X ] Bug Report
[ ] Feature Request
```

## Description

I use one of my components as a wrapper for bootstrap. One of the bootstrap sass files contains the following scss.

```
.blockquote-footer {
  display: block;
  font-size: 80%; // back to default font-size
  color: $blockquote-small-color;

  &::before {
    content: "\2014 \00A0"; // em dash, nbsp
  }
}
```

When running the packagr I get the following error: 

> BUILD ERROR
Octal literal in strict mode (464:14) in D:\Work\Java.git\ng-packaged\.ng_build\ng-packaged\out\foo\foo.component.js
SyntaxError: Octal literal in strict mode (464:14) in D:\Work\Java.git\ng-packaged\.ng_build\ng-packaged\out\foo\foo.component.js

#### How To Reproduce

_for bugs: Please checkout this repo for reproduction:

https://github.com/appienvanveen/ng-packaged

#### Expected Behaviour

Perhaps the unicode should be escaped?

#### Version Information

```
ng-packagr: 2.0.0-rc.7
@angular/*: 5.1.1
typescript: ~2.5.3
rxjs: ^5.4.2",
node: 6.11.0
npm/yarn: 3.10.10
bootstrap 4.0.0-beta.2
```

PR #453, opened 2017-12-27
Added replace line to escape unicode characters

appienvanveen:
Added *.iml to .gitignore

## I'm submitting a...

```
[x ] Bug Fix
[ ] Feature
[x ] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

- other: extended gitignore
- fix: 
https://github.com/dherges/ng-packagr/issues/425

## Description
See: 
https://github.com/dherges/ng-packagr/issues/425

```
[ ] Yes
[ x] No
```

### DIFF
--- a/integration/samples/material/baz/baz.component.scss
+++ b/integration/samples/material/baz/baz.component.scss
@@ -2,4 +2,5 @@
 
 .baz {
     color: $variable-defined-by-file-not-processed-by-loader;
+    content: "\2014 \00A0";
 }
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ yarn-error.log*
 /.vs
 /.idea
 
+# IDE files
+*.iml
+
 # Runtime data
 pids
 *.pid
--- a/integration/samples/material/specs/metadata.ts
+++ b/integration/samples/material/specs/metadata.ts
@@ -38,6 +38,9 @@ describe(`@sample/material`, () => {
       it(`should have style with: "color: red"`, () => {
         expect(METADATA['metadata'].BazComponent.decorators[0].arguments[0].styles[0]).to.have.string('color: "red"');
       });
+      it(`should have style with: "content: \\2014 \\00A0"`, () => {
+        expect(METADATA['metadata'].BazComponent.decorators[0].arguments[0].styles[0]).to.have.string('content: "\\2014 \\00A0"');
+      });
     });
 
   });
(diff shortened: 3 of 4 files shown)


appienvanveen:
Hi @dherges,

can you please have a look a this PR? We are using your great tool at our enterprise and looking forward to this fix. If this solution does not suits you, please point the right direction.

dherges:
Hi @appienvanveen,

sorry for my late reply.

Can you add a test that reproduces the original issue and verifies the new correct behaviour?

In the issue, you mentioned a css stylesheet like:

```css
.foo {
  content: "\2014 \00A0"; // em dash, nbsp
}
```

Can you add a component (plus the css file) to one of the `integration/samples/*` libs?

The test should then verify that the component's style contains the correct css rule – currently, the most reliable way to do so is through the metadata json. The test should `JSON.parse(..)` the metadata json file, locate the component, extract the component's style attribute and verify like `expect(stylesheetString).to.contain('/* what you expect here *)'` – there are already test specs like https://github.com/dherges/ng-packagr/blob/master/integration/samples/material/specs/metadata.ts#L38-L40 and you can do similar

dherges:
Code example listing:

https://github.com/dherges/ng-packagr/blob/17b4e0f/integration/samples/material/specs/metadata.ts#L38-L40

Does it work??!?

appienvanveen:
Hi @dherges! Thanks for replying! 

I've extended the baz integrations test with the unicode parsing test. Hopefully I did what you meant. 
Please let me know if it is okay. If not, I will update the code according to your comments.

integration/samples/material/baz/baz.component.scss
```diff
@@ -2,4 +2,5 @@
 
 .baz {
     color: $variable-defined-by-file-not-processed-by-loader;
+    content: "\2014 \00A0";
```
dherges: Isn't that supposed to be valid unicode character in the css?

https://stackoverflow.com/a/10393517

dherges: Ok, got it. The unicode character triggers `SyntaxError: Octal literal in strict mode`. Not so good....

dherges:
For clarification / memo to self, see https://brajeshwar.github.io/entities/

Related to https://github.com/styled-components/styled-components/issues/188

appienvanveen:
When a new (beta) release will be made including this fix?

PR #468, opened 2018-01-04
feat: add ‘preserveWhitespaces’ to tsconfig

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x ] Feature
```

## Checklist

## Description

"preserveWhitespaces"
Whether to remove blank text nodes from compiled templates. This option is true by default.

It is recommended to set this to explicitly to false as it emits smaller template factory modules and might be set to false by default in the future.

```
[ ] Yes
[x] No
```


dherges:
Yes, atm, I'd prefer the programmatic API approach for overriding the tsconfig. We'll let the default for `preserveWhitespaces: true` and then people can override on a per-component (per-module?) basis or through the programmatic API

deebloo:
You can also set it when you bootstrap your module

dherges:
Adding to my previous comment, setting by default to false will also subtetly break existing library builds in the case where a whitespace in the HTML template is relevant to layouting / stylesheet.

mkoczka:
@dherges Is there actually some way to set preserveWhitespaces to false in the library? Setting it directly on component doesn't work for me.

Final build still does the whitespaces included and also app that is using this library doesn't get rid of the library whitespaces.

dherges:
Like this?

https://www.softwarearchitekt.at/post/2017/08/31/clever-white-space-handling-for-better-performance-in-angular-5.aspx

```
@Component({
  selector: 'app-passenger-search',
  templateUrl: './passenger-search.component.html',
  styleUrls: ['./passenger-search.component.css'],
  preserveWhitespaces: false
})
export class PassengerSearchComponent  {
}
```

mkoczka:
I know how to set it, but it doesn't work when I include the final library bundle to my app.

mkoczka:
I know how to set it, but it doesn't work. I can reproduce it on demo ng-packaged repository. I forked it [here](https://github.com/mkoczka/ng-packaged).

in foo.component.ts
```
@Component({
  selector: 'my-foo',
  template: '<h1> foo </h1>',
  styleUrls: ['./foo.component.scss'],
  preserveWhitespaces: false
})
export class FooComponent {
}
```

When I build the library with `npm run build:lib` and include it as a package into fresh angular-cli generated app. It still has the whitespaces in `<h1> foo </h1>` whether I have preserveWhitespaces set to true or false in the application.

dherges:
@mkoczka This pull request is a bad place to discuss the issue. Can you make sure whether `ngc -p <...>` on your source code works as intended and then either open an issue on this project or on the angular github project?

mkoczka:
@dherges sure

Issue #463, opened 2018-01-04
Build Fails When certain third party package is just in node_modules


ahmedGomaa90:
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description
when including ngx-slimscroll package in the project it just fails.
it says that "built my-app" and it actually built, and moved to dist folder, then it starts building again from a new entry point "my-app/node_modules/ngx-slimscroll", then it fails
![image](https://user-images.githubusercontent.com/29007104/34565342-254a78c0-f163-11e7-8d06-aa8723ef29c7.png)

here's my externals config in ng-package.json
`"ngx-slimscroll": "ngx-slimscroll"`

#### How To Reproduce

just create new angular app and add this package "ngx-slimscroll"
here's their repo link: [https://github.com/jkuri/ngx-slimscroll/blob/master/README.md](url)
when checking their repo, you'll find that authors already use ng-packagr.

#### Expected Behaviour

app built without any errors

#### Version Information

```
ng-packagr: 2.0.0-rc.9
@angular/*: 5.0.0
typescript: 2.4.2
node: 7.4.0
npm/yarn: 4.0.5
ngx-slimscroll: 3.6.1
```

PR #471, opened 2018-01-08
fix: add ng-package.json to excluded files (#463)

ahmedGomaa90:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist
fix: add ng-package.json to excluded files (#463)

## Description
* add ng-package.json to excluded files, as any package contains such a file with this name, ng-packagr tries to bundle secondary entry point reference to issue I reported earlier [#463](https://github.com/dherges/ng-packagr/issues/463)
* also I found an issue when comparing two paths without resolving them, which sometimes leads the condition to be false as the two paths have different separators '/' '\\'

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ng-packagr",
-  "version": "2.0.0-rc.9",
+  "version": "2.0.0-rc.10",
   "description": "Compile and package a TypeScript library to Angular Package Format",
   "keywords": [
     "angular",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ng-packagr",
-  "version": "2.0.0-rc.9",
+  "version": "2.0.0-rc.10",
```
dherges: Hi @ahmedGomaa90 , pull requests must not update the version numbers. Can you submit the PR with source code changes only?

dherges:
Hi @ahmedGomaa90 , thanks for making the changes!

Issue #418, opened 2017-12-17
[Feature] Allow for using "downlevelIteration" when downleveling


JoostK:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description

I'd like to compile a library that depends on `downlevelIteration` to be enabled, a compiler flag that was introduced in TS 2.3. This is related to #256 but affects the options that are used during [downleveling ES2015 to ES5](https://github.com/dherges/ng-packagr/blob/be01602c673889019f5e81fd1157daa3699f0d39/src/lib/steps/tsc.ts#L22-L28), given that these options can not currently be changed. I had used a patched `tsconfig.ngc.json` before in which I enabled `downlevelIteration`, however that only affects the ES2015 build, and not the ES5 build for which the option actually matters ;)

#### Expected Behaviour

Possibly, we may be enabled to specify (some of) the TS compiler flags. This is already being discussed in #256 but would only affect ES2015 compilation, whereas `downlevelIteration` only affects the ES5 downleveling step that is hardcoded at the moment.

#### Version Information

```
ng-packagr: 2.0.0-rc.7
```

JoostK:
Once again, I can whip up a PR but only after this week and once we have agreed upon a workable solution for this situation.

dherges:
Is it safe to set `downLevelIteration` compiler flag, when down-levelling the ESM2015 bundle to a ESM5 bundle?

If yes, I'd say we should simply hard-code that compiler flag to the down-levelling build step.

dherges:
By safe I mean: has `downlevelIteration` no effects other than `for ... of` language constructs?

The resulting es5 code is guaranteed to produce the expected behaviour in es5. Is there a use case for `downlevelIteration` in es2015 target?

JoostK:
I thought about that question as well, and I'm not quite sure why TS decided to introduce a flag for this that is disabled by default.

I think the flag only affects ES5 generation. It does have influence on more stuff than just `for ... of`, e.g.

```
const map = new Map();
const values = [...map.values()]
```

Without `downlevelIteration`:
```
const map = new Map();
const values = map.values().slice();
```

With `downlevelIteration`:
```
const map = new Map();
const values = __spread(map.values());
```

The former is incorrect as Map.values returns an iterable which does not have `slice`. I don't think ES2015 code would be affected by this as syntax can just be preserved.

[...]

PR #475, opened 2018-01-09
feat: include downlevelIteration flag for ES5 bundles

JoostK:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

As discussed in #418

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples/core/specs/umd.ts
+++ b/integration/samples/core/specs/umd.ts
@@ -50,5 +50,12 @@ describe(`@sample/core`, () => {
       expect(API.AngularService).to.be.ok;
     });
 
+    it (`should downlevel iteration`, () => {
+      const iterable = function*() { yield* [1, 2, 3]; };
+      const values = API.AngularService.iterableToArray(iterable());
+
+      expect(values).to.deep.equal([1, 2, 3]);
+    });
+
   });
 });
--- a/src/lib/steps/tsc.ts
+++ b/src/lib/steps/tsc.ts
@@ -24,6 +24,7 @@ export async function downlevelWithTsc(inputFile: string, outputFile: string): P
     module: ModuleKind.ES2015,
     allowJs: true,
     sourceMap: true,
+    downlevelIteration: true,
     mapRoot: path.dirname(inputFile)
   };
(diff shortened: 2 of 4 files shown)


integration/samples/core/specs/umd.ts
```diff
@@ -50,5 +50,12 @@ describe(`@sample/core`, () => {
       expect(API.AngularService).to.be.ok;
     });
 
+    it (`should downlevel iteration`, () => {
+      const iterable = function*() { yield* [1, 2, 3]; };
+      const values = API.AngularService.iterableToArray(iterable());
+
+      expect(values).to.deep.equal([1, 2, 3]);
```
dherges: Hm, maybe I did not understand the downlevelIteration flag correct. The test verifies the `iterableToArray(input: Iterable<T>): Array<T>` method's public API. But shouldn't the test rather verify that the `__spread()` tslib helper is used internally (in the es5 bundle)? 

I guess the above test will always success even without downlevelIteration (since the `*.spec.ts` are run on node@6/7/8/9).

src/lib/steps/tsc.ts
```diff
@@ -24,6 +24,7 @@ export async function downlevelWithTsc(inputFile: string, outputFile: string): P
     module: ModuleKind.ES2015,
     allowJs: true,
     sourceMap: true,
+    downlevelIteration: true,
```
dherges: 👍

integration/samples/core/specs/umd.ts
JoostK: I agree this test is up for debate, a textual contains of the source file may fit as well. But actually, the test will currently fail when the `downlevelIteration` flag is not used, I tested that :-)

The reason it fails is that without `downlevelIteration`, the method will have been compiled into `iterable.slice()` and that is an invalid call on the actual generator instance that is passed in the spec. I figured I might as well test the actual behaviour instead of verifying the source output.

dherges: Ok, I trust you here!

Issue #354, opened 2017-12-02
[PR WELCOME] Language level support for library authors


dherges:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Current Behaviour: language level is tied to [es2015](https://github.com/dherges/ng-packagr/blob/master/src/lib/conf/tsconfig.ngc.json#L23-L26)

Desired use case is es2016 / es2017 language level support for authors writing library code with es2016 features.

#### Expected Behaviour

Usage: `ngPackage.lib.languageLevel = "es2017"`

Internal / implementation detail: `tsconfig compilerOptions.lib = [ /*..*/ es2017 /*..*/ ]`

#### Version Information

```
ng-packagr: v2.0.0-rc.x
```

dherges:
See #162

e-cloud:
it seems simple to update the version number. But obviously it's not. otherwise @dherges will already do it.

should we add more configuration into the `ng-package.json`?

dherges:
Closed by #486

PR #486, opened 2018-01-12
feat: add language level support for library authors

sumigoma:
## I'm submitting a...

```
[ ] Bug Fix
[X] Feature
```

## Checklist

## Description

This PR addresses https://github.com/dherges/ng-packagr/issues/354

From the issue:
Current Behaviour: language level is tied to es2015

Desired use case is es2016 / es2017 language level support for authors writing library code with es2016 features.

**Expected Behaviour**
Usage: `ngPackage.lib.languageLevel = "es2017"`

Internal / implementation detail: `tsconfig compilerOptions.lib = [ /*..*/ es2017 /*..*/ ]`

**Version Information**
```
ng-packagr: v2.0.0-rc.x
```

```
[ ] Yes
[X] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,21 @@ For example, the following would use `index.ts` as the secondary entry point:
 }
 ```
 
+##### How do I use es2016 or es2017 features in my TypeScript library?
+
+You can change the TypeScript language level support in tsconfig by also using the `ngPackage` configuration field in your secondary `package.json` and setting the `languageLevel` property in `lib`:
+For example,:
+
+```json
+{
+  "ngPackage": {
+    "lib": {
+      "languageLevel": [ "dom", "es2017" ]
+    }
+  }
+}
+```
+
 #### How to embed assets in CSS?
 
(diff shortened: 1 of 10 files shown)


sumigoma:
@dherges Is there any particular reason why the language level is tied to `es2015` by default? I've left the default `es2015` here and it can be modified through `ng-package.json` now, but it would be nice to bump this. `@angular/cli` has moved on to `lib: [ "dom", "es2017" ]` in its `tsconfig.json`, for example.

dherges:
This is a good question. Angular core packages are still being built with es2015 ... https://github.com/angular/angular/blob/master/packages/tsconfig-build.json#L16

JoostK:
@sumigoma Thanks for your work!

I reckon that internally, Angular does not currently use es2016/es2017 given that it would require all consumers to include polyfills. Angular CLI chooses to include polyfills by default and therefore has a default library configuration of es2017.

romantul:
@sumigoma: Thanks!!! 
@dherges: Thanks for integrating this feature. I'm impatiently awaiting the new RC for being able to use it :thumbsup:

PR #563, opened 2018-01-27
feat: reduce library bundle size by clean-css

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description
At that moment, `css` and `html` within components cannot be optimized by the consumers by using `Webpack`, `Angular Build Optimizer` or any other tool that bundles simply because these optimization have to happen at a library level when the the transformation happens that turns these assets into strings in the `JS` file

Reduce final bundle size by;
- Using `cleanCss` this will remove comments, new lines, extra quotes among others
- Remove new lines in `HTML` templates as these don't have any effect in the browser

Other component libraries are doing the same. Apart from `preserveWhitespaces` (Which in this case the library develop can do)
https://github.com/angular/material2/blob/73bc948faaf4f3a566256e5e733109f9847bb252/src/lib/button/button.ts#L73

they also have the trim breaklines
https://github.com/angular/material2/blob/73bc948faaf4f3a566256e5e733109f9847bb252/tools/package-tools/inline-resources.ts#L53

https://github.com/angular/material2/blob/73bc948faaf4f3a566256e5e733109f9847bb252/tools/package-tools/gulp/build-scss-task.ts#L14

example
```js
    SkeletonComponent.decorators = [
        { type: core.Component, args: [{
                    selector: "obg-skeleton",
                    template: "<h1>Sample</h1>\n\n<p>some text</p>",
                    styles: [":host {\n  display: block; }\n"]
                },] },
    ];
```
```
[ ] Yes
[x] No
```

Closes: https://github.com/dherges/ng-packagr/issues/614

### DIFF
--- a/src/lib/ng-v5/entry-point/resources/stylesheet.transform.ts
+++ b/src/lib/ng-v5/entry-point/resources/stylesheet.transform.ts
@@ -15,7 +15,7 @@ import * as nodeSassTildeImporter from 'node-sass-tilde-importer';
 import * as less from 'less';
 import * as stylus from 'stylus';
 import * as postcssUrl from 'postcss-url';
-import * as postcssComments from 'postcss-discard-comments';
+import * as postCssClean from 'postcss-clean';
 
 export const stylesheetTransform: Transform = transformFromPromise(async graph => {
   log.info(`Rendering Stylesheets`);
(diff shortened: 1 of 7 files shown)


alan-agius4:
@dherges can you please have a look at this as well?

alan-agius4:
When you have some time can you have a look at this? This sill reduce the bundle size of ui libs quite a bit

alan-agius4:
@dherges When you have some time can you have a look at this? This sill reduce the bundle size of ui libs quite a bit

alan-agius4:
@dherges can you please let me know what to do about this? As for me its kinda critical to reduce the bundle size mainly due to css and new lines.

dherges:
I take only the stylesheet minification in, ok? I still think [newlines in html account for whitespace in browser rendering](https://stackoverflow.com/a/588362)

alan-agius4:
Fair point, thanks for sharing that. And doing the push.

src/lib/ng-v5/entry-point/resources/stylesheet.transform.ts
```diff
@@ -57,13 +57,25 @@ async function renderPostCss(filePath: string, cssStyles: string, cssUrl: CssUrl
...
     postCssPlugins.push(postcssUrl({ url: cssUrl }));
   }
 
+  // this is important to be executed post running `postcssUrl`
+  postCssPlugins.push(
+    autoprefixer({ browsers }),
+    postCssClean({
+      level: {
+        2: {
+          specialComments: false
```
dherges: See list of optoimization at css-clean docs: https://github.com/jakubpawlowicz/clean-css#level-2-optimizations

alan-agius4: Yeah those are on by default, just wanted to sdd this additional option which at the moment is done via another plugin.

PR #573, opened 2018-01-31
fix: embed `tslib` helpers in UMD bundles only

alan-agius4:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description
In a nutshell
- tslib should be embeeded in UMD
- tslib should be external for ESM5 modules
- No tslib for ESM2015
- add tslib as a dependency in package,json, this is important as without this the consumer will break unless there is another library that uses tslib

Closes: #371

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ng-v5/entry-point/write-package.transform.ts
+++ b/src/lib/ng-v5/entry-point/write-package.transform.ts
@@ -60,6 +60,21 @@ export async function writePackageJson(entryPoint: NgEntryPoint, binaries: { [ke
     packageJson[fieldName] = binaries[fieldName];
   }
 
+  // read tslib version from `@angular/compiler` so that our tslib
+  // version at least matches that of angular if we use require('tslib').version
+  // it will get what installed and not the minimum version nor if it is a `~` or `^`
+  if (!(packageJson.dependencies && packageJson.dependencies.tslib)) {
+    const { dependencies: angularDependencies = {} } = require('@angular/compiler/package.json');
+    const TSLIB_VERSION = angularDependencies.tslib;
+
+    if (TSLIB_VERSION) {
+      packageJson.dependencies = {
+        ...packageJson.dependencies,
+        tslib: TSLIB_VERSION
+      };
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/lib/ng-v5/entry-point/write-package.transform.ts
```diff
@@ -60,6 +60,18 @@ export async function writePackageJson(entryPoint: NgEntryPoint, binaries: { [ke
     packageJson[fieldName] = binaries[fieldName];
   }
 
+  // read tslib version from our peerDependencies
+  if (!(packageJson.dependencies && packageJson.dependencies.tslib)) {
+    const tslibVersion = fs.readJsonSync(path.resolve('./package.json')).peerDependencies.tslib;
```
dherges: This may break in a user install when trying to read package.json from current working directory.

Can we re-write to utilize node resolution like this?

```ts
const TSICKLE_VERSION = require('tsickle/package.json').version;
```

alan-agius4: Good catch, I'll fix it soon

alan-agius4: I actual had to change it a bit a left an in code comment

src/lib/ng-v5/entry-point/write-package.transform.ts
```diff
@@ -60,6 +60,21 @@ export async function writePackageJson(entryPoint: NgEntryPoint, binaries: { [ke
     packageJson[fieldName] = binaries[fieldName];
   }
 
+  // read tslib version from `@angular/compiler` so that our tslib
+  // version at least matches that of angular if we use require('tslib').version
+  // it will get what installed and not the minimum version nor if it is a `~` or `^`
+  if (!(packageJson.dependencies && packageJson.dependencies.tslib)) {
+    const { dependencies: angularDependencies = {} } = require('@angular/compiler/package.json');
```
dherges: 👍  That's an even better idea to read it from angular compiler who has it as dependency anyway!

Issue #587, opened 2018-02-05
bug: multiple import statements for `tslib`


alan-agius4:
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

In some cases when downleveling you'll end up with 2 import statements for `tslibs` example

```js
import * as tslib_1 from "tslib";
import { __decorate, __metadata } from 'tslib';
```
This not ideal as the first one requires to import the whole library.

#### How To Reproduce
Just run build
#### Expected Behaviour

Only a single import is present

#### Version Information

```
ng-packagr: 2.0.0-rc.13
@angular/*: x.y.z
typescript: x.y.z
rxjs: x.y.z
node: x.y.z
npm/yarn: x.y.z
```

_please include any version information that might be relevant, e.g. other third-party libraries_

PR #588, opened 2018-02-05
fix: multiple import statements for `tslib`

alan-agius4:
Closes: #587

## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Cherry pick tslib methods and avoid having multiple imports for tslib

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/flatten/flatten.ts
+++ b/src/lib/flatten/flatten.ts
@@ -58,7 +58,14 @@ export async function flattenToFesm15(opts: FlattenOpts): Promise<string> {
 export async function flattenToFesm5(opts: FlattenOpts): Promise<string> {
   const destFile = path.resolve(opts.outDir, 'esm5', opts.flatModuleFile + '.js');
 
-  await downlevelWithTsc(opts.entryFile, destFile);
+  await rollupBundleFile({
+    moduleName: opts.esmModuleId,
+    entry: opts.entryFile,
+    format: 'es',
+    dest: destFile,
+    embedded: opts.embedded,
+    transform: downlevelWithTsc
+  });
 
   return destFile;
 }
--- a/src/lib/flatten/tsc.ts
+++ b/src/lib/flatten/tsc.ts
@@ -1,18 +1,14 @@
 import * as path from 'path';
+import { SourceDescription } from 'rollup';
 import { ScriptTarget, ModuleKind, TranspileOutput, transpileModule, CompilerOptions } from 'typescript';
-import { readFile, outputJson, outputFile as fsOutputFile } from 'fs-extra';
 import { debug } from '../util/log';
 
 /**
  * Downlevels a .js file from ES2015 to ES5. Internally, uses `tsc`.
  *
- * @param inputFile
- * @param outputFile
  */
-export async function downlevelWithTsc(inputFile: string, outputFile: string): Promise<void> {
-  debug(`tsc ${inputFile} to ${outputFile}`);
-  const inputBuffer: Buffer = await readFile(inputFile);
-  const input: string = inputBuffer.toString();
+export async function downlevelWithTsc(code: string, filePath: string): Promise<SourceDescription> {
+  debug(`tsc ${filePath}`);
   const compilerOptions: CompilerOptions = {
     target: ScriptTarget.ES5,
     module: ModuleKind.ES2015,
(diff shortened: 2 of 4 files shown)


src/lib/flatten/flatten.ts
```diff
@@ -57,8 +57,18 @@ export async function flattenToFesm15(opts: FlattenOpts): Promise<string> {
 
 export async function flattenToFesm5(opts: FlattenOpts): Promise<string> {
   const destFile = path.resolve(opts.outDir, 'esm5', opts.flatModuleFile + '.js');
+  // a temp file is required because of the sourcemaps remapping
+  const destTempFile = path.resolve(opts.outDir, 'esm5-downlevel-temp', opts.flatModuleFile + '.js');
 
-  await downlevelWithTsc(opts.entryFile, destFile);
+  await downlevelWithTsc(opts.entryFile, destTempFile);
+  // the below is required in order to avoid multiple imports for tslib
+  await rollupBundleFile({
```
dherges: Hm, I'd prefer to do the down-levelling then in a rollup plugin

For example with bublé : https://github.com/rollup/rollup-plugin-buble

Or register the tsc downlevelling as a transform rollup plugin https://github.com/rollup/rollup/wiki/Plugins#creating-plugins:

```ts
rollup.plugins = [
   ...existinPlugins,
   {
      transform: (source, id ) => {
         const transpiled = ts.transpileModule(/* ... */);

         return {
           code: transpiled.outputText,
           map: transpiled.sourceMap
         };
      }
   }
]
```

alan-agius4: It's cleaner I agree

dherges: Should I update your PR?

alan-agius4: I updated my PR with some code changes as suggested. However, If you need to do some more changes. Feel free.

dherges: Oh, you already did!

src/lib/flatten/tsc.ts
```diff
@@ -1,48 +1,29 @@
...
-export async function downlevelWithTsc(inputFile: string, outputFile: string): Promise<void> {
-  debug(`tsc ${inputFile} to ${outputFile}`);
-  const inputBuffer: Buffer = await readFile(inputFile);
-  const input: string = inputBuffer.toString();
+export async function downlevelWithTsc(code: string, filePath: string): Promise<SourceDescription> {
+  debug(`tsc ${filePath}`);
   const compilerOptions: CompilerOptions = {
     target: ScriptTarget.ES5,
     module: ModuleKind.ES2015,
     allowJs: true,
```
alan-agius4: I am not sure, why this is needed, in general, libraries that generate `.d.ts` cannot use this.

dherges: What do you mean "this"?

dherges: I am not sure what you refer to with "this". Is it the sourcemaps? I remember that was needed as it was transpiling the fesm15 file and w/o stripping it had two "// sourceMappingUrl" lines at end of fesm5 file.

alan-agius4: This was just as a side note. `allowJs` usually shouldn't be enabled. for libraries as libraries usually generate `d.ts` with `declaration: true` and this flag doesn't work with `declaration: true`.

So in theory I don't see the point of having `allowJs` in a library that generated `d.ts`

dherges: `allowJs` is for adding `.js` files to the compilation as we're down-levelling from a `*.js` file (not a `*.ts.`)

> Allow JavaScript files to be compiled.

https://www.typescriptlang.org/docs/handbook/compiler-options.html

alan-agius4: Ah thanks mate for the info

src/lib/flatten/tsc.ts
```diff
@@ -1,48 +1,29 @@
...
-  ]);
+  return {
+    code: transpiled.outputText,
+    map: transpiled.sourceMapText
+  };
 }
-
-const REGEXP = /\/\/# sourceMappingURL=.*\.js\.map/;
-const trimSourceMap = (fileContent: string): string => {
-  if (fileContent.match(REGEXP)) {
```
dherges: Did you check the sourcemaps generated in the dist folder?  There have been issues in the past when an incorrect source map was going to fail the consumer's build.

I am going to double-check later that we're "safe" here

alan-agius4:
They look good to me, but to be honest. I don't know what was the problem.
So, if you could check that would be great

src/lib/flatten/tsc.ts
alan-agius4: They look good to me, but to be honest. I don't know what was the problem.
So, if you could check that would be great

Link to the artifacts;
https://1301-91553810-gh.circle-artifacts.com/0/integration/samples/core/dist/esm5/sample-core.js.map

alan-agius4: any update please?

dherges: I am looking at some source map files with https://sokra.github.io/source-map-visualization/#custom

the ones from the jsx sample look good!

dherges: material also. let' merge!

PR #625, opened 2018-02-16
fix: pass 'setParentNodes' when calling 'createCompilerHost'

deadalusai:
## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

This change passes 'setParentNodes' when calling 'createCompilerHost' as part of the "synthesized" compiler host wrapper.

This resolves an issue we have encountered with `compiler-cli` when it processes `SourceFile` objects
created by this wrapped CompilerHost, caused by the `parent` reference not being set on all AST nodes in the `SourceFile`:

```
TypeError: Cannot read property 'text' of undefined
    at TokenObject.TokenOrIdentifierObject.getText (node_modules\typescript\lib\typescript.js:90769:57)
    at Evaluator.evaluateNode (node_modules\@angular\compiler-cli\src\metadata\evaluator.js:604:66)
    ...
```

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/lib/ts/synthesized-compiler-host.ts
+++ b/src/lib/ts/synthesized-compiler-host.ts
@@ -13,6 +13,7 @@ export function createCompilerHostForSynthesizedSourceFiles(
   sourceFiles: ts.SourceFile[],
   compilerOptions: ts.CompilerOptions
 ): ts.CompilerHost {
+  // FIX(#625): pass `setParentNodes` to the "synthesized" compiler host
   const wrapped = ts.createCompilerHost(compilerOptions, /* setParentNodes */ true);
 
   return {
--- a/src/lib/ts/synthesized-compiler-host.ts
+++ b/src/lib/ts/synthesized-compiler-host.ts
@@ -13,7 +13,7 @@ export function createCompilerHostForSynthesizedSourceFiles(
   sourceFiles: ts.SourceFile[],
   compilerOptions: ts.CompilerOptions
 ): ts.CompilerHost {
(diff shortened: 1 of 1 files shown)


deadalusai:
I couldn't find any other tests already covering this function. I'm happy to add some tests - just point me to where they should go.

dherges:
Hi @deadalusai,

is it possible to reproduce the issue by creaitng a source file with the TypeScrpt APIs like `ts.createSourceFile("foo.ts", "/* source code here */)`?

If yes, put it to the unit tests in the src folder. Any file with a `*.spec.ts` will be covered by tests.

If no, you can try to re-reproduce a full example in `integration/samples`, add a new folder there and put a minimal reproduction (only those source files that trigger the error).

deadalusai:
Hi @dherges. As far as I can tell the branch of code which was causing us a problem is almost never hit in your integration tests. The branch is in `synthesized-compiler-hosts.ts`:

```
36: } else {
37:    return wrapped.getSourceFile(fileName, version);
38: }
```

This branch is hit in exactly one test (when compiling a React component) when `compiler-cli` queries for a `.ts` file (which fails) and then a `.tsx` file (which succeeds).

Is the search of `sourceFiles` for the requested source file expected to **always** succeed?I think passing `setParentNodes: true` is the correct thing to do, however testing that code branch looks like it will be difficult to do.

(We're using a temporary fork of `compiler-cli` while we try to improve its support for barrel files - we hit this issue in ng-packagr while testing that)

dherges:
@deadalusai Ok, i left a notice in the code.

Issue #632, opened 2018-02-20
Provide option to keep the dist folder


alan-agius4:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description
At the moment with every build the destination folder for the distributed is deleted. Which make it almost impossible to create a watch using ng-packagr. As the `npm link` will be lost when the `dist` folder is deleted, and thus you'll end up with webpack giving a error that the module cannot be resolved.

@dherges, if you are okay with implementing this, I can do this feature :)

#### How To Reproduce

- Link a library built with `ng-packagr` ie the dist folder and run a build using `angular-cli` or any other build system

- Re build the library with ng-packagr

#### Expected Behaviour
No error

```
ng-packagr: 2.0.0
```

dherges:
Yes, sure. How will you going to solve it? Keep the folder and truncate its content?

Otherwise it may signal the need for more commands like `ng-packagr clean`, `ng-packagr build`, ...

alan-agius4:
Actually I think truncating the content will still cause an error in Webpack. I need to reconfirm this though. 

My initial idea was to actually just override the files when this option is true.

alan-agius4:
Following, what I mentioned yesterday I did some digging into it and `webpack` doesn't actually trigger a build when a new file is added, but rather than the contents of the file has changed.

So truncating the a folder contents and than copying them won't trigger a build.

I'll open an issue with Webpack maybe it's a bug or there is an option to enable this somehow.

JanneHarju:
I would like to have this also. My scenario is to run concurrenlty ng build --watch with scss-bundle with -w parameter. But because ng-packagr is always removing dist folder I must use wait-on and wait that librarys main d.ts file is generated and after that I start scss-bundle with -w parameter. If I don't wait that main file, my scss-bundle file would be removed also. This scenario also reguires me to clean dist folder before everything. So it is like this: "clean dist" && concurrently "ng build --watch" "scss-bundle some-parameters-here -w folder". By adding wait-on && before scss-bundle I can handle the initial dist clean which would clean my scss-bundle file. But if I could disable initial clean I could remove that wait-on on clean my build process a bit. Same kind of parameter is also in angular cli with basic application so you should ask how they have done https://github.com/angular/angular-cli/wiki/build --delete-output-path

PR #655, opened 2018-03-03
feat: add `deleteOutputPath` option

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description
Provide an option for users to not delete output path before build. The name was choosen for feature parity with NG CLI.

Closes: #632

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/ng-package.schema.json
+++ b/src/ng-package.schema.json
@@ -8,6 +8,11 @@
     "$schema": {
       "type": "string"
     },
+    "deleteDestPath": {
+      "description": "Delete output path before build.",
+      "type": "boolean",
+      "default": true
+    },
     "dest": {
       "description":
         "Destination folder where distributable binaries of the Angular library are written (default: `dist`).",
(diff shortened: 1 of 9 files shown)


src/ng-package.schema.json
```diff
@@ -8,6 +8,11 @@
     "$schema": {
       "type": "string"
     },
+    "deleteOutputPath": {
+      "description": "Delete output path before build.",
+      "type": "boolean",
+      "default": true
```
dherges: Hi, looks good to me! Shouldn't we call it `deleteDestPath` since it's related to the `dest` path?

alan-agius4: I don’t mind renaming it. Originaly I choose that name due to Ng CLI

alan-agius4: Pushed the rename

PR #676, opened 2018-03-08
fix: AMD module ID in UMD output

### LINKED ISSUE #673: Need to give names to the AMD modules nested in UMD bundles
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

Currently a package published by ng-packagr uses anonymous AMD modules:
https://unpkg.com/@toxicable/lite-router@0.0.3/bundles/lite-router.umd.js
`typeof define === 'function' && define.amd ? define(['exports', '@angular/core', 'rxjs/ReplaySubject', 'rxjs/BehaviorSubject', '@angular/router', '@angular/common'], factory) :
	(factory((global['lite-router'] = {}),global.ng.core,global.Rx,global.Rx,global.ng.router,global.ng.common));`

the fact that the first argument to `define` is not a string literal is what tells AMD (eg. requirejs) that the module is unnamed.

That means that libraries produced by ng-packagr must *always* be loaded from URI's or the loader must be configured with a map saying what the name of this module is.

Under bazel's `ts_devserver` we use a faster on-the-fly assembly which is basically just `cat`, and this relies on having named AMD modules.

We should get this in before Angular v6.

kylecordes:
closes #673

Set the UMD module ID per the Rollup API:

https://rollupjs.org/guide/en#exports

## Tests?

The test set up for ng-packagr is relatively complex, I studied it for a few minutes and it was not clear where to add the test for this. With a pointer or two though, I'm certain I could do so.

NOT a breaking change; does not break existing tests, and adding an AMD module ID should not interfere with typical anonymous use of the AMD define inside a UMD.

### DIFF
--- a/src/lib/flatten/rollup.ts
+++ b/src/lib/flatten/rollup.ts
@@ -74,6 +74,7 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void> {
   // Output the bundle to disk
   await bundle.write({
     name: `${opts.moduleName}`,
+    amd: { id: opts.moduleName },
     file: opts.dest,
     format: opts.format,
     banner: '',


kylecordes:
Moot, @thelgevold PR is further along.

PR #685, opened 2018-03-11
feat: facilitate intra-package dependencies between secondary entry points (experimental)

### LINKED ISSUE #652: Determine build graph of entry points
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

- [x] Extract @DavidParks8's algorithm for build groups and dependency graph: https://github.com/dherges/ng-packagr/pull/419#issuecomment-354015908
- [x] Mark `entryPoint.dependsOn(otherEntryPoint)` here: https://github.com/dherges/ng-packagr/blob/master/src/lib/ng-v5/init/analyse-sources.transform.ts#L40-L43
- [x] Extract a `scheduleEntryPoints` transforms from https://github.com/dherges/ng-packagr/blob/master/src/lib/ng-v5/package.transform.ts#L61-L66
- [x] Replace the naive `concat(...array)` with a `concat(...buildGroups)`
- [x] Update the tsconfig used for compilation, add paths mapping to dist folder of other entry points: https://github.com/dherges/ng-packagr/blob/master/src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts#L11-L17

#### Related Issues

#379 #419 

#### Version Information

```
ng-packagr: x.y.z
@angular/*: x.y.z
typescript: x.y.z
rxjs: x.y.z
node: x.y.z
npm/yarn: x.y.z
```

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Enables dependencies between entry points. Suppose you have the following source code:

```
|- package.json (primary, `@my/package`)
   |- shared
       |- package.json (i.e. `@my/package/shared`)
       |- index.ts
   |- feature-a
       |- package.json (i.e. `@my/package/feature-a`)
       |- index.ts
   |- feature-b
       |- package.json (i.e. `@my/package/feature-a`)
       |- index.ts
```

In `./shared/index.ts`:
```ts
export const SHARED_FEATURE = `Running in a loop`;
```

In `./feature-a/index.ts`:
```ts
import { SHARED_FEATURE } from '@my/package/shared';

export const FEATURE_A = `Yeah, ${SHARED_FEATURE}`;
```

In `./feature-b/index.ts`:
```ts
import { FEATURE_A } from '@my/package/feature-a';

export const FEATURE_B = `${FEATURE_A}, fast!`;
```

Closes #652, #379, #419 

## Kudos

Shouts out to @DavidParks8 for the backing algorithm of "furthest distance from root node"!  🦃 👏 

```
[ ] Yes
[x] No
```


dherges:
TODOs:

- [ ] UMD bundles are not correct: feature-a imports from `global.shared` (should be `global.sample.secondary.shared)`
- [ ] Add integration test specs
- [ ] Build consumer app depending on `@sample/secondary/shared`, and so on

matheo:
Wow thanks @dherges !
I'm going to test this branch ASAP.

dherges:
On that note, I'd like to add a consumer app that resolves to the UMD bundles. Who knows how to set up one? Can it be done with `tsc` or should it be "just a html"?

src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
```diff
@@ -1,20 +1,35 @@
...
+      tsConfig.options.paths = {};
+    }
+
+    for (let dep of entryPointDeps) {
+      const depModuleId = dep.data.entryPoint.moduleId;
+
+      if (!tsConfig.options.paths[depModuleId]) {
+        tsConfig.options.paths[depModuleId] = [];
+      }
+      tsConfig.options.paths[depModuleId].push(dep.data.es2015EntryFile);
```
vtabary: Hi all,

After some debugging, I found out that Typescript doesn't like when you leave the ".js" extension file in its paths.

So something like this, will work to use a JS file directly but it won't let typescript find the matching declaration files :
```
"paths": {
  "@sample/secondary/shared": [
    ".ng_pkg_build/sample/out/shared.js"
  ]
}
```

Instead, we should get something like this :
```
"paths": {
  "@sample/secondary/shared": [
    ".ng_pkg_build/sample/out/shared"
  ]
}
```

This will allow Typescript to find the matching declaration files.

So, I wrote something like :
```
import * as path from 'path';
...
const depFilePath = path.resolve(
  path.dirname(dep.data.es2015EntryFile),
  path.basename(dep.data.es2015EntryFile, '.js')
);

tsConfig.options.paths[depModuleId].push(depFilePath);
```

By the way, everything else is working fine on my project.

Hope it will help

matheo:
@dherges can I ask when this is gonna be released?

epelc:
@dherges can one of the sub packages depend on the main package?

ie `@my/package/feature-a` imports `@my/package`?

Having trouble with this. Maybe I'll have some luck if I move into a `@my/package/shared`

JoostK:
@epelc That is absolutely possible! As long as there's no cyclic dependency you can import from any package, provided that they are within the same entrypoint.

epelc:
@JoostK I'm having trouble with this. Getting this error whenever I create a sub package that depends on the primary package. It works fine if I build the primary package before creating the sub package. I assume it's trying to build in the wrong order.

```
Cannot read property 'isSkipSelf' of null
TypeError: Cannot read property 'isSkipSelf' of null
```

Wouldn't it working when the primary package was already built before mean that I don't have cyclic dependencies? I'm not importing any of the sub packages from the primary one

PR #687, opened 2018-03-11
fix: validate non-peerDependencies at build time

dherges:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Publish npm packages with `dependencies` and/or `devDependencies` easily leads to installing multiple versions of a dependency to an application's `node_modules` folder. While this is a desirable solution on server-side or standalone programs, it's a source for bugs on front-end build stacks and UI technologies – you don't want to install two different versions of Angular, don't you?!

A motivation [why `peerDependencies` are the preferred solution for libraries](https://blog.domenic.me/peer-dependencies/) can be found on the Hidden Variables blog (the article is also cross-posted on the npm blog).

With this change, ng-packagr will verify `dependencies` and `devDependencies` at build time and fail the build, if the dependency is not whitelisted explicitly.

```
[ ] Yes
[x] No
```

Technically, this will break library builds that used to compile in prior versions of 2.x. However, the distribution-ready binaries were non-healthy to the ecosystem. As such, this change should be considered a fix to prevent you from bugs rather than a breakage of your build.

### DIFF
--- a/src/lib/ng-package-format/package.ts
+++ b/src/lib/ng-package-format/package.ts
@@ -73,8 +73,10 @@ export class NgPackage {
   public get whitelistedNonPeerDependencies(): string[] {
     // XX: default array values from JSON schema not recognized
     const defValue = ['tslib'];
+    const value = (this.primary.$get('whitelistedNonPeerDependencies') as string[]) || defValue;
 
-    return this.primary.$get('whitelistedNonPeerDependencies') || defValue;
+    // Always append 'tslib' and dedupe
+    return value.concat('tslib').filter((value, index, self) => self.indexOf(value) === index);
   }
 
   private absolutePathFromPrimary(key: string) {
--- a/src/lib/ng-package-format/package.ts
+++ b/src/lib/ng-package-format/package.ts
@@ -70,6 +70,13 @@ export class NgPackage {
     return this.primary.$get('keepLifecycleScripts') === true;
   }
 
+  public get whitelistedNonPeerDependencies(): string[] {
+    // XX: default array values from JSON schema not recognized
+    const defValue = ['tslib'];
+
+    return this.primary.$get('whitelistedNonPeerDependencies') || defValue;
+  }
+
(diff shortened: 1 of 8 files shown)


dherges:
TODO:

- [ ] write an integration test that fails at build time, report the build failure as test success

src/lib/ng-package-format/package.ts
```diff
@@ -66,6 +66,13 @@ export class NgPackage {
     return this.absolutePathFromPrimary('workingDirectory');
   }
 
+  public get whitelistedNonPeerDependencies(): string[] {
+    // XX: default array values from JSON schema not recognized
+    const defValue = ['tslib'];
+
+    return this.primary.$get('whitelistedNonPeerDependencies') || defValue;
```
JoostK: This way, if you specify `whitelistedNonPeerDependencies` yourself, you will also have to include `tslib` yourself. Wouldn't it be easier to just always include `tslib` instead?

dherges: Good question. Is `tslib` necessarily required? Other way round: is it possible to build a package that does not depend on `tslib`?

dherges: The case would be: "I, as library author, wrote a package that does not depend on `tslib` and I'd like to verify that I don't accidentally install it via `dependencies`"

But we always add `tslib` regardless of it's being required or not. So we'd need to check the ESM bundles for an import (via AST to be reliable) to determine that.

JoostK: @dherges hmm, not sure, either choice is somewhat arbitrary. In the spirit of keeping it simple I wouldn't go to great lengths in establishing whether tslib is actually used or not.

vanslly:
The take away for me with this change is that if your library has `dependencies` then they need to ideally be moved to `devDependencies` and added to `peerDependencies`, to firstly avoid nested `node_modules` in packages and to warn about version mismatches.

PR #715, opened 2018-03-24
feat: also create package .tgz

filipesilva:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description
This PR causes `ng-packagr` to also create a `.tgz` file for the library being packaged.

`npm install path/to/directory` creates symlinks since `npm@5.0.0` (see the feature summary in https://github.com/npm/npm/releases/tag/v5.0.0).

The symlink will cause library dependencies to be resolved to in the library project instead of the consuming application.

This is problematic with peer dependencies in general and especially Angular libraries, since the application and the library are using different instances of `@angular/core`. This in turn causes hard to debug errors like the infamous `is not an NgModule` errors.

The correct way to install built packages is to install from the .tgz file. This will not create a symlink and instead install the package as if it was from the npm registry, resolving all dependencies in the consumed project.

```
[ ] Yes
[x] No
```


felicete:
Won't it be better to generate the `.tgz` file inside `./dist` folder?

filipesilva:
@felicete that's a good question, and one that I've thought a while about.

The `.tgz` should be the contents of `dist/`. If we create it, then put it inside, it's a bit odd because it's now modifying the structure it's supposed to contain.

Putting it outside, side by side with the folder, also makes it easier to install (one less dir to go in) and allows you to unpackage it to create a lib folder in the right place.

When you have multiple projects the folder will probably not be just `dist/`. It will be something like `dist/my-lib`. In this case you will have `dist/my-lib.tgz` and `dist/my-lib/package.json`, etc. If it was inside the package directory, it would now be `dist/my-lib/my-lib.tgz`. 

I think putting it outside makes it clearer, more consistent, and easier to use overall.

alan-agius4:
I am currently migrating ng-packagr to APF 6 and I think this might cause a bit of a problem when having multiple entry points.

When we have `dist/something` the `/something` here is a secondary entry point which in angular terms it is like having `/http` under `@angular/common`, that means that with APF 5 and 6 there won't be any JS files under `/something` . as they would be one level up, in their respective ES module folder example: `fesm5`, `esm5` etc... 

Thus I believe that for each `ng-packagr` build there should be a single `tgz` file. Which shall contain all files for that particular library.

alan-agius4:
Yes indeed.

Though I still have a bit of an issue if you do something like this
```ts
	await ngPackagr()
		.forProject("./ng-package.js")
		.withTsConfig("../../tsconfig.build.json")
		.build();

	await bundleScss();
```

As than the `tgz` file won't contain the files modified by the `bundleScss` that said I am not sure if this can be done with `withProviders` which I didn't manage to use it.

Maybe

Issue #705, opened 2018-03-20
Angular Package Format (APF) v6.0


alan-agius4:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description
With a new major Angular version comes a new package format. Some details can be found here though it's not fully updated: https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview#heading=h.k0mh3o8u5hx

#### What changes?
An example build with `APF v6.0` can be found here: https://github.com/angular/core-builds

- Ship both `fesm` and `esm`. 
- Several new entry point in `package.json`. 
```json
  "main": "./bundles/core.umd.js",
  "module": "./fesm5/core.js",
  "es2015": "./fesm2015/core.js",
  "esm5": "./esm5/core.js",
  "esm2015": "./esm2015/core.js",
  "fesm5": "./fesm5/core.js",
  "fesm2015": "./fesm2015/core.js",
  "typings": "./core.d.ts"
```
#### What needs to change in `ng-packagr`?
- The biggest issue to switching from v5 to v6 is that `ng-packagr`, uses a lot of rollup plugins, all these features based on rollup will no longer work. 

`Rollup` should purely be used to generate the `FESM` and `UMD` bundles.

- Downleveling should happen on non flattened files and down outside of rollup.
- Add more entry points for bundlers.
- Ship `ESM5` and `ESM2015` (Unflattened)

#### Expected
Align with `APF v6.0`. 

I still need to look into which bundlers use which entry points in `package.json`, From what I know is that `webpack v4` does tree-shaking on non flat modules `ESM`, but for dev-mode it's best to use `FESM` as it's more performant when having many modules.

elesueur:
This may be related or not, but I am wondering how (if possible at all) to try and build a library that will work with multiple versions of Angular. Currently, I have ng-packagr version 2 which is producing a library using the 5.2 compiler-cli which produces AOT metadata with version 4.

The library doesn't work if I try to load it into an Angular 4 app, complaining of a metadata version mismatch.

There is nothing crazy in this library that would actually stop it from working with Angular 4 as far as I can tell, and it would be nice to be able to support Angular 4,5,6 with one version of the library. In fact, if I go into the metadata file and change the version back to 3, it works just fine with Angular 4 AFAICT.

[...]

PR #597, opened 2018-02-06
test: strip minimal samples for comments integration testing

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
[x] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

- [x] Commit Messages follow the [Conventional Commits](https://conventionalcommits.org/) pattern
  - A feature commit message is prefixed "feat:"
  - A bugfix commit message is prefixed "fix:"
- [ ] Tests for the changes have been added

## Description

Strips integration testing samples to be more minimal.

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #738, opened 2018-03-29
 feat: build libaries in Angular Package Format (APF) v6.0

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description
This commits brings `Angular Package Format` V6 in `ng-packagr`

In a nutshell these are the changes:
- Remove the working directory
- Ship `ESM5`, `ESM2015` modules
- Remove most of `rollup` plugins
- Remapping of `sourcemaps` is not a seperate transformation step
- `tgz` file is created only once and not per entrypoint

BREAKING CHANGES:
There were some important changes, mainly related to `rollup`, most of the options and functionality provided by `rollup` plugins have been removed, due to the fact this is not inline with `APF V6` as one will end up with different outputs in different modules.

- `comments`  option from configuration has been removed
- `licensePath` option from configuration has been removed
- `embedded` was dropped as it increased the chance of having 2 copies of the same library. Only with Angular Package Format 6. This only worked with `UMD` bundles

More info: https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview

Closes: #705

```
[X] Yes
[ ] No
```


JoostK:
Does this mark the start of 3.0? 😃

alan-agius4:
Just a small update. Had a short convo today with Alex Eagle that Igor Minar from Angular pointed out that `tslib` should not be embedded in `UMD` because of you will end up with multiple copies of the embedded library in the same application This has been rolled out with Angular V6, and no libs will be embedded in the UMD bundle

alan-agius4:
The only difference is the for `ESM5` and `ESM2015` the sourcemaps are not inlined. Maybe Igor can explain the pros/cons why they should be inlined?

Thanks @filipesilva for taking the time in reviewing this :)

filipesilva:
I had a chat with @dherges yesterday about this PR, and for him it seems that some of the breaking changes are problematic. Specifically, the removal of `embedded` breaks one of his main usecases.

dherges:
We'll try to find alternatives for `embedded`. The use case is reproduced in the ["typings" example](https://github.com/dherges/ng-packagr/tree/master/integration/samples/typings)

Current ideas are:

 - either maintain the existing behaviour [by `bundledDependencies` npm feature](https://docs.npmjs.com/files/package.json#bundleddependencies)
 - or drop the `embedded` behaviour altogether (requiring users to change their build set up, e.g. code extraction/generation to a `./local` source file rather than embedding from `node_nodules/local`)

eromano:
Guys for the comments? Do we need to remove it? There is a way that can I contribute on this PR adding an alternative comments filter?

alan-agius4:
Click `unsubsribe` right hand side of the screen

alan-agius4:
@dherges, I do understand that this breaking. What I don't understand is the benefit of having embedded/bundled dependencies.

**Pros**
- Doing npm install will be faster as npm doesn't need to do the checksum and resolve the latest package version etc...

**Cons**
- Increase in distributable bundle size since it will bundle the bundledDependency and all it's dependencies
- Can have multiple copied of the same library
- Can have different versions of the same library. Ex: if Lib 1 depends on `lodash: ^4.X.X` either by a `dependency` or even better a `peerDependency` and Lib 2 on  `lodash: ^4.X.X` with a bundledDependencies and Lib 2 was publish a month ago and in the meantime lodash released a new release, the consumer of these libraries will have 2 different versions of lodash.
- Lose npm package hoisting

filipesilva:
@alan-agius4 the usecase that breaks is packages that depend on proprietary libraries (non ng-packagr ones) that are not available outside of an org. So the package must include/embed these libraries directly. It is not a matter of npm install speed proper.

Also I think @eromano was not asking about how to stop receiving comments from this thread. I think he was asking about the `comments` options being removed. Why is that btw? I understand that `embedded` did imply different content in UMD bundles, but `comments`/`license` seem like they would apply to all bundles.

eromano:
@alan-agius4 mqybe I wasn't clear I was talking about the ***comments*** option that you are going to remove with this PR and I have introduced. This is needed in the component library ADF of Alfresco that currently use ng-packagr

alan-agius4:
@eromano apologies for the misunderstanding `comments`/`license` will only work for `UMD`, `FESM5` and `FESM2015` and it was decided to remove them as this plugins no longer work the latest versions of `rollup`. Check the comment by @dherges here: https://github.com/dherges/ng-packagr/issues/705#issuecomment-375957491

eromano:
It is a really big performances issues, leave all the license header in big projects will increase a lot the size of the bundles

eromano:
Can we build something internal? Or your prefer to not add so many concern in ng-packagr?

filipesilva:
@alan-agius4 I hadn't seen that issue yet, thanks for highlighting it!

alan-agius4:
While I do agree that the distributing size will be bigger, it should not have any performance issue for the consumer, as if they use webpack for instance comments will be stripped out if you build in prod mode. 

Also if you take a look at `angular/angular` distributables even they have multiple license headers with the `FESM` files https://unpkg.com/@angular/common@6.0.0-rc.1/fesm2015/common.js

I do agree that it's nicer that there is a single license header, but really there is no performance issue here.

filipesilva:
I'm not a big fan of removing license headers and such in distribution libraries. I think that is a bundling concern, not a packaging concern. The exception is UMD since that is essentially pre-bundled.

alan-agius4:
@filipesilva Indeed though just noticed that in `Angular` they are still there is in UMD https://unpkg.com/@angular/common@6.0.0-rc.1/bundles/common.umd.min.js

That might be a bit of a problem, mainly for the `min` version. Maybe this is a missing uglify rule

In `ng-packagr` license headers are dropped in the min version of the UMD bundle.

alan-agius4:
@filipesilva & @dherges I drafted something related to the `bundledDependencies`. 
https://github.com/alan-agius4/ng-packagr/pull/2

filipesilva:
@alan-agius4 yesterday I did a POC of how moving from `embedded` to `bundleDependencies` would look like: https://github.com/filipesilva/ng-packagr-bundle-deps

Things that are worth mentioning:
- the current `embedded` does not take into account transitive dependencies, so if A depends on B, you need to list both A and B in `embedded`.
- `bundledDependencies` will correctly bundle transitive dependencies
- for `bundledDependencies` to work at all, it needs to have be have the dependencies installed inside the created package and then packaged via `npm pack`
- this last point complicates package creation

Looking at your POC I think it also suffers from the transitive dependencies problem.

alan-agius4:
Yeah you are right, that is a problem when bundling with rollup because of the following line https://github.com/alan-agius4/ng-packagr/pull/2/files#diff-6121a899448ba77704f7281a36157373R11

Re transitive, I did a small update, it's not the cleanest yet but it seems to address it but it's sort of a concept https://github.com/alan-agius4/ng-packagr/pull/2/commits/a55dbe4d12d4574adbb41e96b7c8c32c7a3e0128

alan-agius4:
Yeah you are right, that is a problem when bundling with rollup because of the following line https://github.com/alan-agius4/ng-packagr/pull/2/files#diff-6121a899448ba77704f7281a36157373R11

Re transitive, I did a small update, it's not the cleanest yet but it seems to address it but it's sort of a concept https://github.com/alan-agius4/ng-packagr/pull/2/commits/bc692a2dcb5881c3c3465c0baebcb06dced4e30c

alan-agius4:
@filipesilva in your POC, how do you see the UMDs working with bundledDependencies? If it's used in `node` it will work fine due the module resolution strategy that node has, but what if it's used in a browser or SystemJS for instance?

IgorMinar:
for UMDs you'd have to include the bundled dependencies within the UMDs published. This is what we do for @angular/language-service which needs to be published as a self contained single file.

alan-agius4:
@filipesilva / @IgorMinar, what are the benefits of having sourcemaps inlined for esms? Is it to improve perf of webpack sourcemaps resolution?

Re UMDs, i do a agree that it has to be as a single file, I tried to address it in my POC, by reading and transersing the package.json of the dependencies, Although not sure if there might be a better way to do it. Possible it can done also via TS AST, requiring the embedded module and transvering the source imports statements and building an embedded list from with that.

filipesilva:
@alan-agius4 I do not have an answer to the sourcemaps question myself. Perhaps @IgorMinar does.

Wouldn't rollup embed the needed libs in UMD by not considering them external?

alan-agius4:
Yeah, but by default we are considering all node_modules as externals
unless there are provided in the embeded list. And as highlighted
previously this is a problem if you embed package A, which depends on
package B which depends on C. As you’d need to provide all these deps in
the embedded field. An option can be that you have 2 seperate rollup
builds.

1) You start by bundling the ‘bundledDependency’, treating all paths as
internal so they are all inlined in a single file, ie package A, B and C.
2) Bundle actual library bundle and resolve the embedded paxkage to the
previusly created bundle so you’ll end up with a single file.

Needs a poc but cannot see why this should work.

alan-agius4:
Sorry about the spamming but thanks to your comment @filipesilva, I think I though a way to solve the issue of the transitive dependencies for UMD. 

What if when using rollup, rather than treath everything in node_modules as external, we do the opposite for UMDs everything is embeed unless they are listed in dependencies which are not listed as bundledDependencies and PeerDependencies. This will automatically embed all transitive dependencies of bundledDependencies and keep peer and non bundled dependencies as external.

dherges:
Ok, I try to recap:

1. @IgorMinar says that `bundledDependencies` affects the UMD bundle files: source chunks need to be written to the UMD bundle. For (F)ESM, `bundledDependencies` are not written to the bundle files, the (consumer's) module loader will resolve the `import '...'`.
2. ng-packagr used to have `lib.external` and now `lib.embedded` to explicitly force what source chunks are written to the UMD/FESM bundles by rollup. Plus, an [auto-discovery strategy](https://github.com/dherges/ng-packagr/blob/master/src/lib/flatten/external-module-id-strategy.spec.ts).
3. Shouldn't we remove `lib.embedded` as @alan-agius4 suggests?

- For ESM: no handling for `bundledDependencies`, leave it to the dev to `npm pack` a correct tarball
- For FESM: (same)
- For UMD: handle `bundledDependencies` and include source chunks in the UMD bundle file

Yes, I think that's going to work.

dherges:
Regarding transitive dependencies, i try to make an exmaple:

```json
{
  "name": "foo",
  "bundledDependencies": ["bar"],
  "peerDependencies": {"@angular/core": ".."}
}
```

```json
{
  "name": "bar",
  "dependencies": {"foobar": "..."},
  "peerDependencies": {"@angular/core": ".."}
}
```

What happens when I run `npm pack` for "foo"? "\@angular/core" should not be included because it's peerDependencies.

- Is "bar" included in foo's tarball? Yes, because it's a bundledDependencies
- Is "foobar" included in foo's tarball? From my understanding: no, because the consumer's `npm install` would install "foobar" to `node_modules/foobar` alongside `node_modules/foo` ❓

alan-agius4:
@dherges re the transitive dependencies, from what I understand, it's as you are explaining.

For the transitive deps in `UMD` module I did a sort of auto-discovery (If we take your naming) https://github.com/alan-agius4/ng-packagr/pull/2/commits/178524424bb18cd6e420e3b4be1d6960853ed9e6

What it does basically it's quite straight forward
- `FESM` every `node_module` is external because we real on bundlers to resolve the import
- `UMD` - by default everything is embeeded, unless it's `whitelisted` as a `dependency` or `peerDependency`. In case a `dependency` is listed as a `bundledDependency` the later will super seed the other.

filipesilva:
@dherges in your last example, `bar`, `bar` and `foobar` would be included in the tarball. `@angular/core` would not be in the tarball because it is not in neither dependencies nor bundledDependencies of any of those packages.

The tarball, created through `npm pack`, includes a `node_modules` folder with `foo`, `bar` and `foobar` but not anything else that was in the original `node_modules`.

I like @alan-agius4's strategy of treating everything as non-external for UMDs by default. I think it makes a lot of sense.

And yes, now that I know `npm pack` does way more than just creating a tgz (it also contains `bundledDependencies` logic) I think what I implemented is insuficient. 

There are a couple of extra steps when using `bundledDependencies` and `npm pack` though, and those don't seem trivial to explain or implement, so I don't think there's a big rush to do the whole `npm pack` journey. To be clear, I am talking about how `npm install` must be ran inside the generated package folder before `npm pack` will bundle dependencies. Relative paths to other tarballs also need to be adjusted for `npm install` to work.

It's probably enough to explain to users how to proceed when `bundledDependencies` is used, then gauge demand to see if it's worth it to implement.

filipesilva:
`npm pack` will package the folder with what is there. If the dependencies to be bundled are not there, they will not be part of the bundle. 

At least this is what I saw from testing in https://github.com/filipesilva/ng-packagr-bundle-deps... but maybe this was because I did not have the private dependency listed directly, and installed it from the `.tgz` with `--no-save`. Perhaps if the dependency is listed that is not a problem. Then only the path adjustment is necessary.

alan-agius4:
@filipesilva, you are right. I re-checked and I can confirm that `npm pack` won't install the dependencies.

src/lib/ts/downlevel-transformer.ts
```diff
@@ -0,0 +1,69 @@
...
+    ? Promise.reject(new Error(formatDiagnostics(emitResult.diagnostics, compilerHost)))
+    : Promise.resolve();
+}
+
+/**
+ * Downlevels a .js file from `ES2015` to `ES5`. Internally, uses `tsc`.
+ *
+ * Required for some external as they contains `ES2015` syntax such as `const` and `let`
+ */
+export function downlevelCodeWithTsc(code: string, filePath: string): Promise<{ code: string; map: string }> {
```
dherges: Why two different downlevel methods? I didn't spot it.

src/ng-package.schema.json
```diff
@@ -55,29 +55,12 @@
...
-        "comments": {
-          "description": "Use \"all\" to keep all, \"none\" to remove all the comments or regexes",
-          "type": "string",
-          "default": "all"
-        },
-        "licensePath": {
-          "description": "Licesnse file path to prepend to the bundles",
-          "type": "string",
-          "default": ""
-        },
```
dherges: Note: all three breaking changes are good imo. It's required for moving to APF 6 now. The functionality has to be restored later in v3. v2 will continue to keep that features.

src/lib/ts/downlevel-transformer.ts
alan-agius4: One is required for rollup, which downlevel a single file at a time. The other one is used to downlevel ESM2015 to ESM5 which downlevels all files from an entry point, using a `program`

dherges:
@eromano I think we have to drop comments and license support in v3 - For the moment. The functionality has to be restored after the APF 6 changes. v3 can then also address the issues ,mentioned in #620 and #597. The functioanlity will continue to work in v2.

src/lib/ts/downlevel-transformer.ts
dherges: Ok, why? ngc compiles to esm2015. Then, we donwlevel to esm5. FESM15 takes esm2015 as input. FESM5 takes esm5 as input. Why another downlevelling?

alan-agius4: This is needed for UMD bundles, when we do embedding of external dependencies. This is due the fact that if we are embedding something with ES6 syntax we need to downlevel it. As Rollup won't do that. 

More info
https://github.com/rollup/rollup/issues/2093

SerkanSipahi:
Please consider this issue (https://github.com/dherges/ng-packagr/issues/752) as well as too! The new version of uglify-js (>= 3.3.18), which received by the dependency tree of the package.json from ng-packagr will break the build process (ng-packagr) (as well in this feature branch)!

alan-agius4:
@SerkanSipahi the issue mentioned is not related in the PR, or this feature. Thus, it will be address in another PR.

alan-agius4:
@dherges & @filipesilva I pushed some changed related to the UMD bundles

src/lib/ts/downlevel-transformer.ts
alan-agius4: In all fairness I am considering form es2015 to es5 rather than doenleving Use ngc directly. This will also enable us to have inlinedSourceMaps for es

filipesilva:
@dherges WDYT?

alan-agius4:
@IgorMinar would appreciate some insites about the reason why sourcemaps are inlined now in APF 6.0 for esm modules. Is it to improve performance in webpack?

integration/samples/externals/testing/public_api.ts
```diff
@@ -0,0 +1,5 @@
+import { result } from '@sample/externals';
+import template from 'lodash.template';
+
+const compiled = template('hello <%= user %>!', null, null);
+export const result_test = result;
```
dherges: Note: example code for a bundled depency (prior "embedded")

integration/samples/externals/specs/umd.ts
```diff
@@ -38,16 +35,58 @@ describe(`@sample/externals`, () => {
...
+      expect(BUNDLE).to.not.contain("require('lodash.templatesettings')");
+    });
+
+    it(`should embed transitive 'templateSettings' object`, () => {
+      expect(BUNDLE).to.contain('var templateSettings = {');
+    });
+
+    it(`should embed 'template' function`, () => {
+      expect(BUNDLE).to.contain('template(string, options, guard)');
+    });
```
dherges: Note: test specs for a bundled depency (prior "embedded")

dherges: bundled depencencies are embedded in UMD bundles. This test verifies the behaviour.

integration/samples/externals/specs/fesm2015.ts
```diff
@@ -0,0 +1,46 @@
...
+        encoding: 'utf-8'
+      });
+    });
+
+    it(`should import '@sample/externals`, () => {
+      expect(BUNDLE).to.contain("from '@sample/externals'");
+    });
+
+    it(`should import 'lodash.template`, () => {
+      expect(BUNDLE).to.contain("from 'lodash.template'");
```
dherges: Note: test spec for a bundled depency (prior "embedded").

In EcmaScript modules (both FESM and ESM), the bundled dependency is not embedded. The `import ... from ...` will be resolved to a sub-folder in `<library-name>/node_modules/<bundled-dependency-name>`

integration/samples/externals/package.json
```diff
@@ -5,9 +5,17 @@
...
-    "rxjs": "^5.4.2"
+    "rxjs": "^5.4.2",
+    "@angular/core": "^4.1.2",
+    "@angular/common": "^4.1.2",
+    "@angular/cdk": "5.2.4"
+  },
+  "bundledDependencies": ["lodash.template"],
+  "dependencies": {
+    "lodash.template": "*"
   },
```
dherges: Note: Usage example for a bundled depency (prior "embedded")

src/lib/flatten/external-module-id-strategy.spec.ts
```diff
@@ -1,34 +1,143 @@
...
+          expect(externalModuleIdStrategy.isExternalDependency('/foo/bar')).to.be.false;
+        });
+
+        it(`should return 'false' for absolute paths`, () => {
+          expect(externalModuleIdStrategy.isExternalDependency(__filename)).to.be.false;
+        });
+
+        it(`should return 'false' for external modules like '@angular/core'`, () => {
+          expect(externalModuleIdStrategy.isExternalDependency('@angular/core')).to.be.false;
+        });
```
dherges: ```it(`should return 'false' for a bundled dependency)```

src/lib/ng-package-format/entry-point.ts
```diff
@@ -55,6 +55,30 @@ export class NgEntryPoint {
...
+
+    return {
+      metadata: pathJoinWithDest(secondaryDir, `${flatModuleFile}.metadata.json`),
+      declarations: pathJoinWithDest(secondaryDir, `${flatModuleFile}.d.ts`),
+      esm2015: pathJoinWithDest('esm2015', secondaryDir, `${flatModuleFile}.js`),
+      esm5: pathJoinWithDest('esm5', secondaryDir, `${flatModuleFile}.js`),
+      fesm2015: pathJoinWithDest('fesm2015', `${flatModuleFile}.js`),
+      fesm5: pathJoinWithDest('fesm5', `${flatModuleFile}.js`),
+      umd: pathJoinWithDest('bundles', `${flatModuleFile}.umd.js`),
+      umdMinified: pathJoinWithDest('bundles', `${flatModuleFile}.umd.min.js`)
```
dherges: :+1:

src/lib/ng-package-format/package.ts
```diff
@@ -61,7 +61,9 @@ export class NgPackage {
     return this.absolutePathFromPrimary('dest');
   }
 
-  /** Absolute path of the package's working directory (used for intermediate file storage). */
+  /** Absolute path of the package's working directory (used for intermediate file storage).
+   * @deprecated This is unused and will be removed in the next major version
+   */
   public get workingDirectory(): DirectoryPath {
     return this.absolutePathFromPrimary('workingDirectory');
```
dherges: :+1:

dherges:
@filipesilva I went through the changes and looks good!

alan-agius4:
Thanks @dherges, I bet now we can go through some deprecations and clean them up?

PR #769, opened 2018-04-16
feat: update tsickle to =>0.27.3, rollup to ^0.57.1 

### LINKED ISSUE #679: ng-packagr generates weird constructor for subclass which breaks angular dependency injection
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description

Creating a subclass with an implicit constructor and initialized property causes the esm5 codegen to emit a weird form of the "pass-through" constructor.

#### How To Reproduce

1. Create a new empty library and prep it for `ng-packagr`
2. Add the following code:

    ```typescript
    export class Dependency { }
    export class Base {
        constructor (public dependency: Dependency) { }
    }
    export class SubBad extends Base {
        // This class generates a constructor which uses `super.apply(this, __spread(arguments))`
        property = 'value'
    }
    export class SubOk extends Base {
        // This class generates a constructor which uses `super.apply(this, arguments)`
    }
    ```
3. Run `ng-packagr`

4. Observe that the ems5 bundle file contains following JS for `SubBad` and `SubOk`:

    ```javascript
    var SubBad = /** @class */ (function (_super) {
        tslib_1.__extends(SubBad, _super);
        function SubBad() {
            var _this = _super.apply(this, tslib_1.__spread(arguments)) || this;
            // This class generates a constructor which uses `super.apply(this, __spread(arguments))`
            _this.property = 'value';
            return _this;
        }
        return SubBad;
    }(Base));
    var SubOk = /** @class */ (function (_super) {
        tslib_1.__extends(SubOk, _super);
        function SubOk() {
            return _super !== null && _super.apply(this, arguments) || this;
        }
        return SubOk;
    }(Base));
    ```

5. Note the difference in the way the `super.apply` call is generated.

#### Expected Behaviour

The `super.apply` call should be the same for both constructors. This is currently necessary because Angular uses a regular expression (!) to determine if a constructor function is a pass-through, and the version which uses `tslib_1.__spread` fails this check. This bug https://github.com/dherges/ng-packagr/issues/631 then takes effect and Angular fails to construct the type correctly.

The regex check in Angular: https://github.com/angular/angular/blob/2fee5cc0958e08d16d6e549a056f3289138d5739/packages/core/src/reflection/reflection_capabilities.ts#L71

Angular should probably improve its code for detecting pass-through constructors, however it seems to be a non-trivial problem (it is hard to tell the difference between an zero-param constructor and a generated pass-through constructor). 

We could extend the regex in angular, however after passing through webpack the constructor is rewritten to look more like this:

```javascript
function SubBad () {
    var _this = _super.apply(this, __WEBPACK_IMPORTED_MODULE_0_tslib__["__spread"](arguments)) || this;
    _this.bordered = false;
    return _this;
}
```

...so maintaining the regex long-term seems like a losing battle.

#### Version Information

```
ng-packagr:            2.2.0
@angular/compiler:     5.2.8
@angular/compiler-cli: 5.2.8
rollup:                0.55.5
tsickle:               0.27.2
typescript:            2.7.2
```

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Updates tsickle to version >= 0.27.3

Resolves #679

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "postcss-url": "^7.3.0",
     "read-pkg-up": "^3.0.0",
     "rimraf": "^2.6.1",
-    "rollup": "^0.55.0",
+    "rollup": "^0.57.1",
     "rollup-plugin-commonjs": "8.3.0",
     "rollup-plugin-node-resolve": "^3.0.0",
     "rxjs": "^5.5.0",
@@ -57,7 +57,7 @@
   "peerDependencies": {
     "@angular/compiler": "^5.0.0 || ^6.0.0-rc.0",
     "@angular/compiler-cli": "^5.0.0 || ^6.0.0-rc.0",
-    "tsickle": ">=0.25.5",
+    "tsickle": ">=0.27.3",
     "tslib": "^1.7.1",
     "typescript": ">=2.4.2"
   },
@@ -101,9 +101,10 @@
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "sinon": "^4.0.0",
+    "source-map": "^0.6.1",
     "standard-version": "^4.0.0",
     "ts-node": "^5.0.0",
-    "tsickle": "^0.27.1",
+    "tsickle": "^0.27.3",
     "tslib": "^1.7.1",
     "typescript": "~2.6.1",
     "zone.js": "^0.8.10"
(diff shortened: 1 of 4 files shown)


dherges:
Out of nothing :zap:, the above compilation error occurred from `RawSourceMap` type from `source-map`.

Updated the source base to [return a `RawSourceMap` instead of a plain `string` (was handled by rollup before)](https://github.com/rollup/rollup/blob/master/src/utils/transformBundle.ts#L28-L31)

By that, also updated rollup to latest version.

Issue #785, opened 2018-04-19
Inline sourcemaps vs external sourcemaps in esm5 and esm2015 APF6


alan-agius4:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description
At the moment in  RC 1 of `ng-package` V3. We are shipping the sourcemaps as external files in `esm5` and `esm2015` folder, Angular shipped these as inlined. https://unpkg.com/@angular/common@6.0.0-rc.5/esm5/

From what I understand this is to speed up the build time of the application to reduce the amount of required file resolutions. Though I'd like to know if there are more reason for this, mainly from @IgorMinar as I think he was the master mind behind this :)

Shall we go with the same approach?

What will bring?
 - Instead of downleveling `TS`, from esm2015 to esm5 `NGC` will be used.
 - No need to re-wire sourcemaps for esm5 and esm2015.

//cc @filipesilva

filipesilva:
I personally am not aware of any benefit to either approach (internal or external sourcemaps).

alan-agius4:
The only thing I can think and based on the code is that for large projects it’s more efficient to use inline since webpack will have less file to resolve and read.

Basically, for `FESM` you have a single `js` file with a single `map` which is super fast to resolve. But for `ESM` you can potentially have `100s` of `js` files. which when sourcemaps are not inlined you would have that amount `x2` IO operations as webpack would have to resolve and do IO operations for both the `js` and `map` files. Rather than parsing a single file and get the code and sourcemap it will have to parse a file, resolve and read the sourcemap, which in large projects will slow down the build quite a lot.

This I think is also something similar why some people choose to flatten their `scss` files into a single one. (Similar to Angular Material 2)

As in general less IO operations is equal to faster builds.

In addition seeing the APF document (https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview#), it looks like for `webpack 4` using `FESM` is something interim until a size problem is resolved.

> As of webpack v4 the flattening of ES modules optimization should not be necessary for webpack users, and in fact theoretically we should be able to get better code-splitting without flattening of modules in webpack, but in practice we still see size regressions when using unflattened modules as input for webpack v4. This is why "module" and "es2015" package.json entries still point to fesm files. We are investigating this issue and expect that we'll switch the "module" and "es2015" package.json entry-points to unflattened files when the size regression issue is resolved.

[...]

Issue #803, opened 2018-04-26
Sourcemaps are not working when rebuilding during webpack watch


peterbakonyi05:
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description
When setting `deleteDestPath` to `false` and linking the `dist` folder of a package to an application, sourcemaps are working fine (ts files are visible) during the first build. When I do a change in the original source and rebuild the library, webpack watch will pick up the `esm5.js` files instead of the `.ts` as source map.

The problem seems to be the 2-step writing of sourcemaps because when `fesm5` JS file is emitted and `webpack` picks up the change, the sourcemaps are still pointing to the `esm5.js` files and they get updated later during the build. Probably a temporary folder or virtual files are needed to emit the changes in one step to the `dist` folder so that when `webpack` picks up the change, it reads the final source maps.

#### Expected Behaviour
When rebuilding the library, the TS source maps are picked up.

#### Version Information

```
$ node_modules/.bin/ng-packagr --version
ng-packagr: 3.0.0-rc.1
@angular/*: 5.2.9
typescript: 2.7.2
rxjs: 5.5.8
node: 8
npm: 5.6.0
```

PR #812, opened 2018-04-27
fix: sourcemaps are not working when rebuilding during webpack watch

alan-agius4:
## I'm submitting a...

```
[x] Bug Fix
[x] Feature
```

## Checklist

## Description

What this change brings is;
- `relocation` and `rewiring` of sourcemaps happens during flattening part.
- `ESM5` and `ESM2015` has inlined sourcemaps which conforms with `APF V6`
- Downleveling of `ESM2015` to `ESM5` has been removed in favor of an `ngc` parallel build this **improve the build time by approx 10%** in some cases. **This is also a step towards incremental builds**. 

If you compare the `yarn test` of this CI and the another `yarn test` in another CI it takes approx **60 seconds less** on 17 very small samples.

Closes: #785, #803 

```
[ ] Yes
[x] No
```


src/lib/ng-v5/entry-point/write-bundles.transform.ts
```diff
@@ -67,7 +68,7 @@ async function writeFlatBundleFiles(destinationFiles: DestinationFiles, opts: Fl
   log.info('Bundling to UMD');
   await flattenToUmd({
     ...opts,
-    entryFile: fesm5,
+    entryFile: esm5,
```
alan-agius4: this is needed to keep the sourcemaps intact

dherges:
This changes sourcemaps from external `.map` files to inline source maps. Is that right?

alan-agius4:
For esm and esm2015 yes, for the others they are still external

alan-agius4:
bump

alan-agius4:
fixed conflicts from upstream

dherges:
I think the fesm/umd bundles are missing the last line `//# sourceMappingURL=<filePath>.js.map`

src/lib/flatten/rollup.ts
```diff
@@ -49,13 +52,29 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void> {
...
-  await bundle.write({
-    name: `${opts.moduleName}`,
-    file: opts.dest,
+  const result = await bundle.generate({
+    name: opts.moduleName,
     format: opts.format,
     amd: opts.amd,
     banner: '',
     globals: moduleId => umdModuleIdStrategy(moduleId, opts.umdModuleIds || {}),
     sourcemap: true
```
dherges: For rollup, we'd need to change the output options:

```diff
diff --git a/src/lib/flatten/rollup.ts b/src/lib/flatten/rollup.ts
index 67ee77b..1dc459f 100644
--- a/src/lib/flatten/rollup.ts
+++ b/src/lib/flatten/rollup.ts@@ -52,13 +52,16 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void[]> {
   });

   // Output the bundle to disk
+  const sourcemapFile = `${opts.dest}.map`;
   const result = await bundle.generate({
+    file: opts.dest,
     name: opts.moduleName,
     format: opts.format,
     amd: opts.amd,
     banner: '',
     globals: moduleId => umdModuleIdStrategy(moduleId, opts.umdModuleIds || {}),-    sourcemap: true
+    sourcemap: 'inline',
+    sourcemapFile
   });

   // relocate sourcemaps
@@ -76,5 +79,5 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void[]> {     }
   });

-  return Promise.all([outputJson(`${opts.dest}.map`, result.map), outputFile(opts.dest, result.code)]);
+  return Promise.all([outputJson(sourcemapFile, result.map), outputFile(opts.dest, result.code)]); }
```


Still then, we have to append the final line ourselves as [rollup does it in their `build` executable command](https://github.com/rollup/rollup/blob/master/bin/src/run/build.ts#L59-L63)

dherges: See https://github.com/rollup/rollup/issues/121

alan-agius4: Yeah, realised that. Thanks for your help. I’ll do these within the hour

alan-agius4: I pushed an update

src/lib/flatten/rollup.ts
```diff
@@ -83,6 +83,6 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void[]> {
 
   // rollup doesn't add a sourceMappingURL
   // https://github.com/rollup/rollup/issues/121
-  result.code = `${result.code}//# sourceMappingURL=${sourcemapFile}`;
+  result.code = `${result.code}\n//# sourceMappingURL=${sourcemapFile}`;
```
alan-agius4: no need for this :P

src/lib/flatten/rollup.ts
alan-agius4: bdw, kudos for realising that `sourceMappingURL` was not being added

src/lib/flatten/rollup.ts
```diff
@@ -83,6 +83,6 @@ export async function rollupBundleFile(opts: RollupOptions): Promise<void[]> {
 
   // rollup doesn't add a sourceMappingURL
   // https://github.com/rollup/rollup/issues/121
-  result.code = `${result.code}\n//# sourceMappingURL=${sourcemapFile}`;
+  result.code = `${result.code}\n//# sourceMappingURL=${result.map.toUrl()}`;
```
alan-agius4: 👍

alan-agius4:
Thanks

PR #826, opened 2018-05-01
feat: enable `fullTemplateTypeCheck`

alan-agius4:
## I'm submitting a...

```
[x] Bug Fix
[x] Feature
```

## Checklist

## Description

Angular recommends to turn this flag on for packages and libraries

https://github.com/angular/angular/blob/master/packages/compiler/design/separate_compilation.md#simplified-options

https://angular.io/guide/aot-compiler#fulltemplatetypecheck

Also this will solve a lot of issue that people faced recently with `Angular 6` https://github.com/angular/angular/issues/23609

```
[ ] Yes
[ ] No
```


src/lib/ngc/compile-source-files.ts
```diff
@@ -47,8 +47,8 @@ export async function compileSourceFiles(
     rootNames: tsConfig.rootNames,
     options: tsConfigOptions,
     emitCallback: createEmitCallback(tsConfigOptions),
-    host: ngCompilerHost,
-    oldProgram: ngProgram
+    emitFlags: tsConfig.emitFlags,
```
alan-agius4: I had to re-add this as Angular Compiler are not handling all cases properly.
https://github.com/angular/angular/blob/5.2.x/packages/compiler-cli/src/perform_compile.ts#L152

If I don't add this, `ngsummary` files will be generating when adding `fullTemplateTypeCheck` even though `skipTemplateCodegen` is set

I'll try to have a look at it and open an issue with Angular.

src/lib/ts/conf/tsconfig.ngc.json
```diff
@@ -4,7 +4,8 @@
     "flatModuleId": "AUTOGENERATED",
     "flatModuleOutFile": "AUTOGENERATED",
     "skipTemplateCodegen": true,
-    "strictMetadataEmit": true
+    "strictMetadataEmit": true,
+    "fullTemplateTypeCheck": true
```
alan-agius4: JFYI, I found some other bugs related to the `strictMetadataEmit` just in case someone opens an issue.

https://github.com/angular/angular/issues/23629
https://github.com/angular/angular/issues/23609

src/lib/ngc/compile-source-files.ts
alan-agius4: https://github.com/angular/angular/issues/23662

PR #842, opened 2018-05-06
build: update to angular 6 final

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description
update to angular 6 final 

```
[ ] Yes
[x] No
```


dherges:
Ah,it was running `ng build --target development`, `ng build --target production` and `ng build --target production --aot`.

In the beginning, it was more safe to verify both JiT and AoT in combination with the dev/prod build of the CLI (because of the implicites of tree-shaking and build optimizer, etc. etc.) were we wanted to make sure that the generated npm packages do work end to end (from library author to consumer build in all the cases).

We should update the scripts in `consumers/ng-cli/package.json` where possible.

alan-agius4:
I'll take a look at that when I am back home. I also noticed this error 'error upath@1.0.2: The engine "node" is incompatible with this module. Expected version ">=4 <=9".'

alan-agius4:
Ci is green.

PR #845, opened 2018-05-07
fix: create an array large enough to hold all buckets

georgiee:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

- [] Tests for the changes have been added

## Description

This will fix an error that appears only when using multiple secondary entry-points depending on each other.

Updated the examples to cause the error and fixed it with a proper initialization of the involved array holding our buckets. I'm not able to create a test for this scenario and think the samples will cover it properly.

That's the error it will fix:
```
Cannot read property 'push' of undefined
TypeError: Cannot read property 'push' of undefined
    at nodeDepths.forEach (/Users/geokal/job/opensource/ng-packagr/src/lib/brocc/depth.ts:107:22)
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples/secondary/feature-c/feature-c.ts
+++ b/integration/samples/secondary/feature-c/feature-c.ts
@@ -1,0 +1,4 @@
+import { FEATURE_A } from '@sample/secondary/feature-a';
+import { FEATURE_B } from '@sample/secondary/feature-b';
+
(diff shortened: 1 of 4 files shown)


dherges:
@georgiee Perfect, thanks! Will look at it in the next days.

PR #850, opened 2018-05-08
fix: handle nested entry points with same name

alan-agius4:
Using `basepath` causes an issue that nested entrypoints will lose their full path, which might also cause accidental override of modules if they are named the same

## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```


src/tsconfig.specs.json
```diff
@@ -5,23 +5,10 @@
...
-    ],
-    "types": [
-      "chai",
-      "mocha",
-      "node",
-      "sinon"
-    ]
+    "lib": ["es2015", "es2016"],
+    "types": ["chai", "mocha", "node", "sinon"],
+    "typeRoots": ["../node_modules/@types"]
```
alan-agius4: without these tests are failing on windows

src/lib/ng-package-format/entry-point.ts
```diff
@@ -66,7 +66,7 @@ export class NgEntryPoint {
 
     if (this.secondaryData) {
       primaryDestPath = this.secondaryData.primaryDestinationPath;
-      secondaryDir = path.basename(this.secondaryData.destinationPath);
+      secondaryDir = path.relative(primaryDestPath, this.secondaryData.destinationPath);
```
alan-agius4: Oops, if you have something like `core/testing` with basepath it will become `testing`

alan-agius4:
@dherges when you have some time can you have a look at this please?

src/lib/ng-package-format/entry-point.ts
alan-agius4: It was part of the APF as before we had a flat dist structure

Issue #822, opened 2018-04-30
Error during template compile, function calls are not supported in decorators.


Tyler-V:
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description

One of my libraries dependencies requires a .forRoot() call to declare the module with providers.

I am including this Module.forRoot() in my module and it works fine, however running packagr for the project fails on compilation.

#### How To Reproduce

```
import { DeviceDetectorModule } from 'ngx-device-detector';

@NgModule({
    imports: [
        DeviceDetectorModule.forRoot(),
    ],
   ...
}
```

```
BUILD ERROR
Error during template compile of 'EasyGridModule'
  Function calls are not supported in decorators but 'DeviceDetectorModule' was called.

Error: Error during template compile of 'EasyGridModule'
  Function calls are not supported in decorators but 'DeviceDetectorModule' was called.

    at Object.<anonymous> (C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ngc\compile-source-files.js:69:35)
    at Generator.next (<anonymous>)
    at C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ngc\compile-source-files.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ngc\compile-source-files.js:3:12)
    at Object.compileSourceFiles (C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ngc\compile-source-files.js:19:12)
    at Object.<anonymous> (C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ng-v5\entry-point\ts\compile-ngc.transform.js:37:60)
    at Generator.next (<anonymous>)
    at C:\Users\Tyler\Documents\GitHub\@ngx-easy\node_modules\ng-packagr\lib\ng-v5\entry-point\ts\compile-ngc.transform.js:7:71
    at new Promise (<anonymous>)
```

#### Expected Behaviour

Successful compilation for publishing.

#### Version Information

```
    "@angular/compiler-cli": "^6.0.0-rc.1",
    "@angular-devkit/build-ng-packagr": "~0.5.10",
    "@angular-devkit/build-angular": "~0.5.10",
    "ng-packagr": "^2.4.2",
[...]

PR #860, opened 2018-05-10
fix: do not disable fullTemplateTypeCheck when ES5 downleveling

JoostK:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

In #826 the option was enabled in tsconfig.ngc.json and then in #812 the option was disabled again when downleveling to ES5. This then causes #822 to resurface again, which was supposedly fixed by #826.

No integration test has been added at this moment, I'm hoping this can get in during the RCs to get real world feedback if the function calls in decorator bugs are now actually fully resolved. With the changes in this PR applied to 3.0.0-rc4 I have been able to successfully build my project that was also suffering from this issue after upgrading from 2.4 to the 3.0 RC.

## More Information

After some two hour debug session in the guts of Angular's compiler, I finally tracked down angular/angular#23609 to originate from [this piece of code when reading the metadata of a library](https://github.com/angular/angular/blob/117c7eebc388ccc7a138a468d79adcd314630471/packages/compiler/src/aot/static_symbol_resolver.ts#L381-L385). As can be seen there, with certain conditions Angular rejects most of the class metadata such that function calls cannot be resolved, resulting in the error.

Following above condition [we follow the code into `summaryResolver.isLibraryFile`](https://github.com/angular/angular/blob/117c7eebc388ccc7a138a468d79adcd314630471/packages/compiler/src/aot/summary_resolver.ts#L52-L57) to find that it calls into [the compiler host to determine if the file not a source file](https://github.com/angular/angular/blob/117c7eebc388ccc7a138a468d79adcd314630471/packages/compiler-cli/src/transformers/compiler_host.ts#L480-L482) that is the actual culprit. Here we find that `skipTemplateCodegen` and `fullTemplateTypeCheck` ultimately affect the decision to elide some of the metadata.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
+++ b/src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
@@ -61,8 +61,7 @@ export const compileNgcTransform: Transform = transformFromPromise(async graph =
       // the options are here, to improve the build time
       declaration: false,
       declarationDir: undefined,
-      skipMetadataEmit: true,
-      fullTemplateTypeCheck: false
+      skipMetadataEmit: true
     })
   ]);
 


alan-agius4:
Very well explained!

alan-agius4:
Very well explained.

PR #866, opened 2018-05-12
fix: allow sideEffects to be set as an array

mattlewis92:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Currently side effects can only be a boolean, however it is allowed to be an array: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

```
[ ] Yes
[x] No
```


src/lib/ng-package-format/entry-point.ts
```diff
@@ -172,7 +172,9 @@ export class NgEntryPoint {
    *
    * @link https://github.com/webpack/webpack/tree/master/examples/side-effects
    */
-  public get hasSideEffects(): boolean {
-    return this.packageJson['sideEffects'] === true;
+  public get sideEffects(): boolean {
```
alan-agius4: The return type here is incorrect. I should be `boolean | sting[]`

Why not simply?
```ts
return this.packageJson['sideEffects'] || false;
```

mattlewis92: Good catch, fixed 😄

mattlewis92:
@alan-agius4 yes that's a much simpler implementation, changed it now 😄 Thanks for reviewing!

PR #868, opened 2018-05-12
fix: embed tslib helpers in umd bundle

mattlewis92:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

tslib helpers shouldn't be embedded in the UMD bundles. This was how ng-packagr 2.x worked and how angular itself works in all versions (inc 6.x):
https://npmcdn.com/@angular/common@6.0.1/bundles/common.umd.js
https://npmcdn.com/@angular/material@6.0.1/bundles/material.umd.js

If you search for `__` and you will see the helpers embedded and there is no mention of `tslib` in the source. 

I believe the reason for this is to make it easier for system.js users to consume angular. 

Any changes please give me a shout 😄 

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples/core/specs/sourcemaps.ts
+++ b/integration/samples/core/specs/sourcemaps.ts
@@ -44,9 +44,9 @@ describe(`@sample/core`, () => {
     });
 
     it(`should reference each 'sources' path with a common prefix`, () => {
-      const everyUeveryMe = (sourceMap.sources as string[]).every(
-        fileName => fileName.startsWith('ng://@sample/core') && fileName.endsWith('.ts')
-      );
+      const everyUeveryMe = (sourceMap.sources as string[])
+        .filter(fileName => fileName !== 'null')
+        .every(fileName => fileName.startsWith('ng://@sample/core') && fileName.endsWith('.ts'));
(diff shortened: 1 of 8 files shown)


alan-agius4:
Strange, as I remember a convo in a PR with @igorminar and @AlexEagle and they pointed out that `tslib` should not be embedded in UMDs.

integration/samples/core/specs/sourcemaps.ts
```diff
@@ -44,9 +44,9 @@ describe(`@sample/core`, () => {
     });
 
     it(`should reference each 'sources' path with a common prefix`, () => {
-      const everyUeveryMe = (sourceMap.sources as string[]).every(
-        fileName => fileName.startsWith('ng://@sample/core') && fileName.endsWith('.ts')
-      );
+      const everyUeveryMe = (sourceMap.sources as string[])
+        .filter(fileName => fileName !== 'null')
```
alan-agius4: this is odd, why there should a `null` as file name in the sourcemap?

alan-agius4:
Just found a PR which in fact they are now adding embedding it https://github.com/angular/angular/pull/23354

PR #940, opened 2018-06-12
feat: add a option to pass a custom `tsconfig` in `CLI`

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
[ ] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

- [x] Commit Messages follow the [Conventional Commits](https://conventionalcommits.org/) pattern
  - A feature commit message is prefixed "feat:"
  - A bugfix commit message is prefixed "fix:"
- [ ] Tests for the changes have been added

## Description

A new argument has been added so that one can provide a custom tsconfig when using the `CLI`

Example:
`
ng-packagr --config tsconfig.lib.json --p ng-packagr.json
`
## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #956, opened 2018-06-18
feat: update `lib` compiler option to `es2018`

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description
`languageLevel` / `libs` are now currently specially treated as they can be overridden without the needs to override the tsconfig. tsconfig options should be standardised and updated via the overriding a tsconfig. Tsconfig can be also overridden via the cli if this PR gets merged. https://github.com/dherges/ng-packagr/pull/940

The following `peerDependencies` are now required
- `tslib: ^1.9.0`
- `typescript: ^2.7.0`

`languageLevel` option has been dropped from the configuration, in favor of the standard tsconfig option `lib`. The default is now `es2018`. In case this needs to be changed. This can to be done by overriding the tsconfig. More on how to override the tsconfig can be found here: https://github.com/dherges/ng-packagr/blob/master/docs/override-tsconfig.md

```
[X] Yes
[ ] No
```


docs/DESIGN.md
```diff
@@ -1,16 +1,13 @@
-Design Doc: "ng-packagr"
-========================
+# Design Doc: "ng-packagr"
```
alan-agius4: Note: auto-formatting by prettier hook

package.json
```diff
@@ -58,8 +58,8 @@
     "@angular/compiler": "^6.0.0",
     "@angular/compiler-cli": "^6.0.0",
     "tsickle": ">=0.27.3",
-    "tslib": "^1.7.1",
-    "typescript": ">=2.4.2"
+    "tslib": "^1.9.1",
+    "typescript": ">=2.7.2"
```
alan-agius4: Note: TypeScript 2.7.x+ is required for lib es2018

dherges:
Hi all, do we really want to make setting the ECMAScript language level only by tsconfig?

Why am I not so much a fan of tsconfig? There are several combination of options that will make the build fail for Angular libraries (either for the developer build or for the application/consumer build). There are several other options that are mandatory and that need to be overriden by the build. Third and last and most important for me:

The building, bundling and packaging is a complex task. tsconfig is only one small part of that. The issue of node and npm and the js community is that **EVERY**tool brings its own config. We developers spent more time on some details and internals of compilers and transpilers (_and and and_). We should however spent very little time on what the expected result (e.g. "a library in es2018 js") is and spent a lot of our time on how we get there (i.e. writing the code for components and services and so on).

alan-agius4:
I see you your point, and I have updated the PR accordingly.

Though in all fairness what's so different between `lib` (Language Level) and `noFallthroughCasesInSwitch` for instance and one is exposed to be overridden and the other no?

PR #988, opened 2018-07-02
refactor: remove custom delete strategy for filecache

alan-agius4:
We don't need to trigger on directory removed or renamed as chokidar will fire the changes for each file

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,12 +2,17 @@
   "buildOnSave": false,
   "compileOnSave": false,
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es2015",
     "moduleResolution": "node",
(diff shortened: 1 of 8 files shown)


tsconfig.json
```diff
@@ -2,12 +2,17 @@
...
-    "target": "es5",
-    "lib": ["es2015", "es2016"],
-    "types": ["chai", "mocha", "node", "sinon"],
-    "jsx": "react"
+    "noEmitOnError": true,
+    "noImplicitAny": false,
+    "sourceMap": true,
+    "inlineSourceMap": false,
+    "inlineSources": true,
+    "lib": ["es2018"],
```
alan-agius4: Nit: this is required for this change as `lib: es2015` does't contain `Array.includes`

PR #1022, opened 2018-07-26
perf: reduce memory consumption

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description
Use a single file cache to reduce memory consumption when having multiple entrypoints. This also help to improve the build time due to after the analyse the files will be already cached.

```
[ ] Yes
[X] No
```


src/lib/ts/tsconfig.ts
```diff
@@ -99,26 +99,43 @@ export const initializeTsConfig = (defaultTsConfig: TsConfig, entryPoints: Entry
...
+
+  for (let dep of entryPoints) {
+    const { entryPoint } = dep.data;
+    const { moduleId, destinationFiles, entryFilePath } = entryPoint;
+    const mappedPath = [pointToSource ? entryFilePath : destinationFiles.declarations];
+
+    if (!tsConfig.options.paths[moduleId]) {
+      tsConfig.options.paths[moduleId] = mappedPath;
+    } else {
+      tsConfig.options.paths[moduleId].concat(mappedPath).reverse();
```
alan-agius4: Note: we reverse this as it we know the `dist` or `src` thus this paths should be preferred. Otherwise fallback.

Issue #797, opened 2018-04-24
New TSConfig Management


noelmace:
## Type of Issue

```
[ ] Bug Report
[ ] Feature Request
[x] discussion
```

## Description

This discussion is created in order to permit a more global discussion about TSConfig management, customisation and refactoring, directly following https://github.com/dherges/ng-packagr/pull/790#discussion_r183844424.

noelmace:
## current mechanism

***required -> consumer -> override (see #786)***

### required

Was created in order to keep the "right" error message, that is to say the one from TSC, which, in my opinion, reflect the best the reason of the error (at least, better than a "your TSConfig doesn't respect the requirements of ng-packagr").

If we want a better error message, we'll need to write one for each case. 

### override

are options that will break the ng-packagr build entirely (without a readable error, as we could see with all the related issues which was really difficult to trace).

### Why

1. writing a tsconfig could **always** break your build. TSC and NGC really have a better maturity (so, error messages) that ng-packagr on this for much cases
2. but ng-packagr always relies on some of this options in a really specific way, what leads to a set of mandatory options which will have to be overriden in order to hide some misleading erros

### How

The identification of the options and their sorting out between a and b cases, was proof tested  during https://github.com/angular/devkit/pull/750 ...

## Some other (futur) options

*(see options 1 and 2 proposed in #790 (comment) as this is a response to it)*

I personally don't think the overriding is a good long term solution, as it could mislead the user on the behaviour of ng-packagr. This would require some documentation at least (but doc is not fix :wink: ). I consider that this is a temporary solution, waiting for a better one. So, I disagree with the option 1. The option 2, at least, could be a better solution, but it relies again on documentation, and we can be sure that this will lead to misunderstandings.

All this question reflect the global approach of ng-packagr : wrapping a lot of tools instead of making just a new one.

Globally, there is some issues with the 3 ways tsconfig are generated and used, as it's misleading a lot of users when using a custom TSConfig (especially with imports / paths, like https://github.com/dherges/ng-packagr/issues/285#issuecomment-384101460, https://github.com/dherges/ng-packagr/issues/519#issuecomment-383418228 or https://github.com/dherges/ng-packagr/issues/791)

Following that, I personally see two other options, but this would need some more refactoring : 
1. simply hide the TSConfig as well (as it could be misleading), and only permit specific options
    a. by writing a new json schema
    b. or by adding a new typescript options object to ng-package (and removing the withTsConfig method from the public api, as advanced users could always use providers)
[...]

PR #1024, opened 2018-07-27
refactor: better tsconfig defaults and remove code for rewiring

alan-agius4:
refactor: better tsconfig defaults and remove code for rewiring

With this change consumers will be able to remove the below from their tsconfig, and now we can still have the correct emit flags 

```js
"flatModuleId": "AUTOGENERATED",	
"flatModuleOutFile": "AUTOGENERATED",
```

Closes: #797

//cc @filipesilva


src/lib/ts/tsconfig.ts
```diff
@@ -19,7 +17,30 @@ export function readDefaultTsConfig(fileName?: string): TsConfig {
...
+
+    // ng compiler to options
+    enableResourceInlining: true,
+
+    // these are required to set the appropriate EmitFlags
+    flatModuleId: 'AUTOGENERATED',
+    flatModuleOutFile: 'AUTOGENERATED'
+  };
+
+  return ng.readConfiguration(fileName, extraOptions);
```
alan-agius4: Note: By providing the extraOptions when reading the configuration, ngc can appropriate set the correct emitFlags

src/lib/ngc/compile-source-files.ts
```diff
@@ -35,27 +33,16 @@ export async function compileSourceFiles(
...
   const result = ng.performCompilation({
     rootNames: tsConfig.rootNames,
     options: tsConfigOptions,
     host: ngCompilerHost,
     emitCallback: createEmitCallback(tsConfigOptions),
-    emitFlags: tsConfig.emitFlags
+    emitFlags
   });
 
-  // XX(hack): redirect the `*.metadata.json` to the correct outDir
```
alan-agius4: Note: Not needed since we can do this, with `redirect-write-file-compiler-host.ts`

src/lib/ngc/compile-source-files.ts
```diff
@@ -35,27 +33,16 @@ export async function compileSourceFiles(
 
   // Don't use `ng.emit` as it doesn't output all errors.
   // https://github.com/angular/angular/issues/24024
+  const emitFlags = tsConfigOptions.declaration ? tsConfig.emitFlags : ng.EmitFlags.JS;
```
alan-agius4: Note: if there are not declarations to be emitted, we only want to emit the js files, this is due than when compiling in es5 the metadata and dts will be already emitted with the previous transpilation of es2015

src/lib/ts/tsconfig.ts
filipesilva: Does this mechanism allow overwriting `flatModuleId`, `flatModuleOutFile` in the user tsconfig?

alan-agius4: Not really, but it was never the case.

filipesilva: Fair enough 👍

src/lib/ts/redirect-write-file-compiler-host.ts
```diff
@@ -23,7 +23,7 @@ export function redirectWriteFileCompilerHost(
       sourceFiles?: ReadonlyArray<ts.SourceFile>
     ) => {
       let filePath = fileName;
-      if (fileName.endsWith('.d.ts')) {
+      if (/(.d.ts|.metadata.json)$/.test(fileName)) {
```
JoostK: You forgot to escape the dots! 😉

alan-agius4: You are right. Thanks.

src/lib/ts/tsconfig.ts
```diff
@@ -50,55 +71,43 @@ export const initializeTsConfig = (defaultTsConfig: TsConfig, entryPoints: Entry
...
+
+    // Add paths mappings for dependencies
+    const entryPointDeps = entryPoints.filter(x => x.data.entryPoint.moduleId !== entryPoint.moduleId);
+    if (entryPointDeps.length > 0) {
+      if (!tsConfig.options.paths) {
+        tsConfig.options.paths = {};
+      }
+    }
+
     currentEntryPoint.data.tsConfig = tsConfig;
```
dherges: Had merge conflicts here. Are the conflicts resolved properly?

alan-agius4: No, I will fix them now

alan-agius4: @dherges conflicts should be good now, waiting for CI to go green.

Thanks anyways for trying 👍

PR #1026, opened 2018-07-27
Pipe moduleId in umdModuleIdStrategy

douglascross:
The UMD gets generated with null inputs when you don't because that is what rollup does.
This causes co-dependent libraries in angular to fail when used in plain script tags.

## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/lib/flatten/umd-module-id-strategy.spec.ts
+++ b/src/lib/flatten/umd-module-id-strategy.spec.ts
@@ -44,8 +44,8 @@ describe(`rollup`, () => {
       expect(umdModuleIdStrategy('foo', { foo: FOO_MODULE })).to.equal(FOO_MODULE);
     });
 
-    it(`should map 'foo' to '' when no 'umdModuleIds' is provided`, () => {
-      expect(umdModuleIdStrategy('foo')).to.empty;
+    it(`should map 'foo' to 'foo' when no 'umdModuleIds' is provided`, () => {
+      expect(umdModuleIdStrategy('foo')).to.equal('foo');
     });
   });
 });
--- a/src/lib/flatten/umd-module-id-strategy.ts
+++ b/src/lib/flatten/umd-module-id-strategy.ts
@@ -17,17 +17,9 @@ export const umdModuleIdStrategy = (moduleId: string, umdModuleIds: { [key: stri
     return `ng.${regMatch[1]}`.replace(/\//g, '.');
   }
 
-  if (moduleId === 'rxjs') {
-    return 'rxjs';
-  }
-
   if ((regMatch = /^rxjs\/(\/?.*)/.exec(moduleId))) {
     return `rxjs.${regMatch[1]}`;
   }
 
-  if (moduleId === 'tslib') {
-    return 'tslib';
-  }
-
-  return ''; // leave it up to rollup to guess the global name
+  return moduleId; // pipe moduleId because rollup will set to null otherwise.
 };


douglascross:
Changes made.
Needed to take time to confirm my earlier investigation.

rollup has an internal function called globalProp that just returns a null string option if the globalName is empty. And the globalName is empty because of the strategy functions return value of empty.

I did see that you could set umdModuleIds, but why when the id is the same as the name.

douglascross:
I'm unfamiliar with how to create an integration test.

Here are the steps to reproduce the issue. Hope it helps.

* Create an angular application (`ng new my-app` should be fine)
* Create a library `ng g library my-lib-1` and build it `ng build my-lib-1`
* export some test variable in my-lib-1 eg. `export var foo = 'bar'`
* Create a 2nd library `ng g library my-lib-2` (you can give it a peer dependency of 'my-lib-1' but it may not matter) and build it `ng build my-lib-2`
* Have my-lib-2 use my-lib-1
````
    import { foo } from 'my-lib-1';
    console.log(foo);
````
* Make sure the libraries are available somehow to the main application. (Code is back at work, I'm making this comment while at home on my weekend.)
* Modify the index.html to have 2 script tags
````
     <script src="location-moved-to/my-lib-1.umd.js"></script>
     <script src="location-moved-to/my-lib-2.umd.js"></script>
````
* Will result in a failure to reference foo because object is null. The UMD factory is populated with null for globals.

Expected `globals['my-lib-1']` to be passed to the factory in the UMD wrapper instead of `null`

**Comment by alan-agius4** [comment]
I took a deeper look into this, and it looks to be a bug within `rollup` rather than `ng-packagr`. 

We rely on the fact that if the module name is not found, a warning will be displayed to the user, so that it is fixed. If the guessed name is not correct. However, it looks like in some recent version there was a change which breaked this. 

We don't want to use the moduleId as UmdModuleId without a warning because in a lot of cases this won't be correct. As usually these have different values.

I created a PR in rollup to address this issue at source: https://github.com/rollup/rollup/pull/2359

**Comment by alan-agius4** [comment]
I took a deeper look into this, and it looks to be a bug within `rollup` rather than `ng-packagr`. 

We rely on the fact that if the module name is not found, a warning will be displayed to the user so that a potential wrong module name can be fixed, as in a lot of cases the guessed module name will be incorrect. However, it looks like in some recent version there was a change which breaked this, and no warning is being displayed and also the `null` value is being tricked down to the bundle emitted.

Ideally we don't use the `moduleId` as `UmdModuleId` without a warning because in a lot of cases this won't be correct.

I created a PR in rollup to address this issue at source: https://github.com/rollup/rollup/pull/2359

That said, it is highly recommended that you specify the UMD module identifiers yourself:
https://github.com/dherges/ng-packagr/blob/master/docs/dependencies.md#resolving-umd-module-identifiers

**Comment by douglascross** [comment]
That would mean my configs would look like the following...
And I'd have to add this for any dependencies where name matched global.
It just seems a little redundant.

````
{
  "$schema": "../../../src/ng-package.schema.json",
  "lib": {
    "umdModuleIds": {
      "my-lib-1": "my-lib-1",
      "my-lib-2": "my-lib-2"
    }
  }
}
````
See how it seems redundant, when name = value?
What am I not understanding?

Why isn't it okay to assume that the global value matches the name for an AMD or require etc?

**Comment by alan-agius4** [comment]
It is redundant because of the name you choose of your library and in this case indeed `name === value` however, in most cases this is not the cases.

Note: than when the issue in rollup is solved, you can omit your settings and you'll just get a warning stating that the name was guessed.

Some examples when these are not the same,  for instance if you library is scoped, which is quite popular now a days, the name would be `@foo/bar` for instance, and the value should be `foo.bar`, if you use `lodash`, the name would be `lodash` and the value would be `_` as `_` is the correct UMD id for `lodash`.

**Comment by douglascross** [comment]
In those cases you will need to specify, but in my case you do not.
Isn't it better to allow for reasonable defaults to just work?

**Comment by alan-agius4** [comment]
They will work, once the issue at upstream is solved.

**Comment by douglascross** [comment]
... okay so it will start working again but with a warning?

**Comment by alan-agius4** [comment]
Yes, they will work, once the issue at upstream is solved.

**Comment by douglascross** [comment]
I understand now, and fair enough.

I wonder then, would it be worthwhile to come up with a way to avoid the warning when the referenced modules are produced in angular itself?
Because we know that the module name matches the global variable name?

**Comment by alan-agius4** [comment]
That not entirely correct, projects created with Angular CLI don't always have the name matches the global name. For instance when created scoped libraries example;

```bash
ng generate library @myScope/myLib
```

Apart from that no one is stopping users from changing the library name from the `package.json`.

alan-agius4:
Super-seeded by fix at source: 
https://github.com/rollup/rollup/pull/2359

douglascross:
@alan-agius4 thanks for all your help.

PR #1170, opened 2018-12-29
Deps update

alan-agius4:


package.json
```diff
@@ -57,23 +57,21 @@
     "update-notifier": "^2.3.0"
   },
   "peerDependencies": {
-    "@angular/compiler": "^6.0.0 || ^7.0.0-beta.0",
-    "@angular/compiler-cli": "^6.0.0 || ^7.0.0-beta.0",
+    "@angular/compiler": "^6.0.0 || ^7.2.0-beta.0",
```
JoostK: @alan-agius4 Shouldn't this include `^7.0.0`?

alan-agius4: Yeah fixing it as we speak

PR #1191, opened 2019-01-23
fix: use ts.compilerHost as moduleResolutionHost for tsickle.emitWith…

alan-agius4:
Compilation fails with exception of get directoryExists on undefined

TypeError: Cannot read property 'directoryExists' of undefined
    at Object.directoryProbablyExists
Because of moduleResolutionHost is undefined due TsickleHost created does not contain host property
Related lines:
https://github.com/angular/tsickle/blob/master/src/tsickle.ts#L172
https://github.com/Microsoft/TypeScript/blob/0a7c92864d628c02e450f23ab2547ed1f5954b56/src/compiler/utilities.ts#L4019

Fixes: https://github.com/ng-packagr/ng-packagr/issues/1189

### DIFF
--- a/src/lib/ngc/create-emit-callback.ts
+++ b/src/lib/ngc/create-emit-callback.ts
@@ -64,7 +64,7 @@ export function createEmitCallback(options: api.CompilerOptions): api.TsEmitCall
   }) =>
     tsickle.emitWithTsickle(
       program,
-      { ...tsickleHost, options, host },
+      { ...tsickleHost, options, ...{ host } },
       host,
       options,
       targetSourceFile,


JoostK:
Am I missing something, how is this different from before? Looking at the TS emit output they should achieve the same thing?

https://www.typescriptlang.org/play/index.html#src=const%20a%20%3D%20%7B%20...tsickleHost%2C%20options%2C%20...%7B%20host%20%7D%20%7D%3B%0D%0Aconst%20b%20%3D%20%7B%20...tsickleHost%2C%20options%2C%20host%20%7D

PR #1196, opened 2019-01-25
perf: reduce the calls to dependsOn

alan-agius4:
This call is heavy when having a lot of nodes. It has multiple concats and filters and can cause perf issues when having a lot of nodes


JoostK:
I see that `Node.dependsOn` accepts both a single node and a list of nodes to add. Could we separate those out into two separate methods, such that the one that accepts only a single node can do this optimization internally.

alan-agius4:
Good idea, I’ll need to check a bit though.

PR #1213, opened 2019-02-11
feat: remove tsickle from direct peerDepedencies

### LINKED ISSUE #1202: Remove tsickle from direct peerDependencies
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description
Users outside of Google don't usually need closure annotations hence `annotateForClosureCompiler`, should be turned off.

`tsickle` should be removed from the `peerDependencies` and it's usage should be similar to the one of `compiler-cli` https://github.com/angular/angular/blob/94e305f48e9f919f3163a50556f2fe9c517572dc/packages/compiler-cli/src/main.ts#L95

Related to: https://github.com/angular/angular-cli/pull/13546

alan-agius4:
Fix #1202

## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

BREAKING CHANGE: Users outside of Google don't usually need closure annotations hence `annotateForClosureCompiler` is turned off by default. In case users want to emit closure compatable code. they need to install `tsickle` and enable opt it this feature.

```
[x] Yes
[ ] No
```

### DIFF
--- a/integration/samples/core/specs/es5-api.ts
+++ b/integration/samples/core/specs/es5-api.ts
@@ -36,7 +36,7 @@ describe(`@sample/core`, () => {
     });
 
     it(`should import TS helpers from 'tslib'`, () => {
-      expect(BUNDLE).to.contain(`import { __spread } from 'tslib'`);
+      expect(BUNDLE).to.contain('tslib');
     });
 
     it(`should downlevel iteration`, () => {
--- a/integration/samples/core/specs/es2015-api.ts
+++ b/integration/samples/core/specs/es2015-api.ts
@@ -36,8 +36,8 @@ describe(`@sample/core`, () => {
       expect(API.AngularService).to.be.ok;
     });
 
-    it(`should not import TS helpers from 'tslib'`, () => {
-      expect(BUNDLE).not.to.contain('tslib');
+    it(`should import TS helpers from 'tslib'`, () => {
+      expect(BUNDLE).to.contain('tslib');
     });
   });
 });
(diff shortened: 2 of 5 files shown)


integration/samples/core/specs/es5-api.ts
```diff
@@ -36,7 +36,7 @@ describe(`@sample/core`, () => {
     });
 
     it(`should import TS helpers from 'tslib'`, () => {
-      expect(BUNDLE).to.contain(`import { __spread } from 'tslib'`);
+      expect(BUNDLE).to.contain('tslib');
```
JoostK: Can we be more specific in this assertion? I have no idea how tslib is supposed to be used when looking at this assertion.

alan-agius4: What we care of is that tslib is present. We don't really care how it is used as that is up to the tsc compiler. tslib is used by the tsc compiler and it is emittes the helpers based on the code emitted. Previously tsickle did some transformations which didn't require tslib for certain code parts.

Re the above, question. in most cases tslib was already being included and should have for example when using the spread operator, extends etc... It was not because decorator metadata was transformed by tsickle.

tslib as no effect on lib authors and consumers.

ViceIce:
You've forgotten to remove the dependency in the version command.

https://github.com/ng-packagr/ng-packagr/blob/bfbe02b158a609fc3bd3bb72b3befaaba8f05098/src/lib/commands/version.command.ts#L6

This import can now fail and should be catched

alan-agius4:
Yeah, will do a fix soon.

CarstenLeue:
How do I opt-in to annotate for closure compiler?

alan-agius4:
You need to add `annotateForClosureCompiler` in your tsconfig

https://angular.io/guide/aot-compiler#annotateforclosurecompiler

PR #1214, opened 2019-02-11
fix: emit complete diagnostics

alan-agius4:
At the moment we are not emitting information diagnostics and also diagnostics emitted during the compilation

### DIFF
--- a/src/lib/ngc/compile-source-files.ts
+++ b/src/lib/ngc/compile-source-files.ts
@@ -53,14 +53,6 @@ export async function compileSourceFiles(
 
   await ngProgram.loadNgStructureAsync();
 
-  const diagnostics = [
-    ...ngProgram.getNgOptionDiagnostics(),
-    ...ngProgram.getTsSyntacticDiagnostics(),
-    ...ngProgram.getTsSemanticDiagnostics(),
-    ...ngProgram.getNgSemanticDiagnostics(),
-    ...ngProgram.getNgStructuralDiagnostics()
-  ];
-
   log.debug(
     `ngc program structure is reused: ${
       oldProgram ? (oldProgram.getTsProgram() as any).structureIsReused : 'No old program'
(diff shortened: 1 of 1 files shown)


src/lib/ngc/compile-source-files.ts
```diff
@@ -69,13 +71,24 @@ export async function compileSourceFiles(
...
 
-  ngProgram.emit({
-    emitCallback: createEmitCallback(tsConfigOptions),
-    emitFlags
-  });
+  if (allDiagnostics.length === 0) {
+    return;
+  }
+
+  const exitCode = ng.exitCodeFromResult(allDiagnostics);
```
JoostK: I was misled in thinking that this is the same result as on line 64, however new diagnostics may have been added during emit (the whole point of this PR). Can we add some comments to guide the reader (or just me perhaps 🤷‍♂️) what is going on here :)

src/lib/ngc/compile-source-files.ts
```diff
@@ -53,14 +53,16 @@ export async function compileSourceFiles(
...
-  const diagnostics = [
+  const allDiagnostics = [
     ...ngProgram.getNgOptionDiagnostics(),
     ...ngProgram.getTsSyntacticDiagnostics(),
     ...ngProgram.getTsSemanticDiagnostics(),
     ...ngProgram.getNgSemanticDiagnostics(),
     ...ngProgram.getNgStructuralDiagnostics()
   ];
 
+  const hasError = ng.exitCodeFromResult(allDiagnostics) > 0;
```
JoostK: Can we move this closer to where it's being read, and change it into:

```suggestion
  const hasError = ng.exitCodeFromResult(allDiagnostics) !== 0;
```

src/lib/ngc/compile-source-files.ts
alan-agius4: Yeah I’ll add them 😊

Issue #1175, opened 2019-01-02
Angular 8 / Ivy: how is implementation of ng-packagr affected by Ivy/ng 8?


dherges:
## Type of Issue

```
[ ] Bug Report
[ ] Feature Request
```

## Description

To discuss: 

- breaking changes in compiler API?
- expected changes in compiler output

#### How To Reproduce

_tbd_

#### Expected Behaviour

_tbd_

#### Version Information

```
$ node_modules/.bin/ng-packagr --version
ng-packagr: 4.5.x
```

dherges:
Angular Compilr API's used:

- ng.ParsedConfiguration
- ng.readConfiguration()
- ng.CompilerOptions
- ng.CompilerHost
- ng.createCompilerHost()
- ng.Program
- ng.createProgram()
- ng.exitCodeFromResult()
- ng.formatDiagnostics()

dherges:
What would be needed to migrate ng-packagr to the CLI repo?

 - bunch of integration tests (samples libraries in `integration/samples`, plus fixtures): need to migrate to karma (I use mocha, chai)
 - build stack for stylesheet processing would come in (node-sass, less, node-sass-tilde-importer, browserlist, postcss, plus  a few more extra dependencies)
 - peerDependencies need to be reviewed: v4 of ng-packagr accepts Angular ^6.0.0 | ^7.0.0 | ^7.2.0-beta.0 - align major versions of ng-packagr and Angular?
 - custom implementation of a build pipeline and simple graph: possibly duplicated code with Devkit Architect?

JoostK:
@dherges it is perhaps good to know that ngtsc seems to become available as a TypeScript plugin (to be used from the Bazel rules). That will likely mean that we'll just be able to use the native TS API together with that plugin enabled. Regarding the loading of resources we should look into ngtsc's `ResourceLoader` interface. At this moment however I don't think it's completely clear what this will look like exactly.

<sup>Above information is entirely assumption based, I do not know about the exact plans.<sup>

dherges:
Idea to look at: flattening d.ts files:

 - https://api-extractor.com/
 - https://github.com/Microsoft/web-build-tools/tree/master/apps/api-extractor

mnahkies:
@dherges is there any update on this, now that Angular 8 is released and stable?

alan-agius4:
[...]

PR #1228, opened 2019-03-03
Flatten typings, d.ts rollup

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

See #1175 

See angular/angular#29086

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   "main": "./public_api.js",
   "typings": "./public_api.d.ts",
   "dependencies": {
+    "@microsoft/api-extractor": "^6.3.0",
     "@ngtools/json-schema": "^1.1.0",
     "autoprefixer": "^9.0.0",
     "browserslist": "^4.0.0",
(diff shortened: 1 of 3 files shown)


dherges:
Findings:

- for human beings, it would be best to publish the public `.d.ts` from the entry point only
  - does not contain old greeks
  - does not contain `@internal` annotated symbols
- for the compiler, i think we need to publish the internal `.d.ts` from the generated bundle index
  - contains old greeks
  - contains `@internal` symbols

Further tasks:
 - remove single module `.d.ts` files from dest folder
 - store typings in `<dest>/typings` folder for all entry points (similar to `bundles`, `fesm5` folder, and so on)

Nice-To-Have:
 - is it possible to re-use tsc compiler program for extracting API?

package.json
```diff
@@ -27,6 +27,7 @@
   "main": "./public_api.js",
   "typings": "./public_api.d.ts",
   "dependencies": {
+    "@microsoft/api-extractor": "^6.3.0",
```
alan-agius4: Ideally version 7 is used. Version 6 can be challenging to use in existing projects.

alan-agius4:
You also need to re-wire the ngc metadata.

See https://github.com/angular/angular/pull/28588 for more context

alan-agius4:
Running this in watch mode might make incremental builds significantly slower since it cannot used a previous program not the file cache.

The main reason why is the api extractor doesn’t nesserary support the latest or same ts version that the uses have installed.

Issue #1255, opened 2019-04-01
npm run packagr  "Cannot read property ngPackage of Undefined"


SaranVairam:
## Undefined Issue

```
BUILD ERROR
Cannot read property 'ngPackage' of undefined
TypeError: Cannot read property 'ngPackage' of undefined
    at Object.<anonymous> (path\node_module
s\ng-packagr\lib\ng-v5\discover-packages.js:37:20)
    at Generator.next (<anonymous>)
    at fulfilled (path\node_modules\ng-pack
agr\lib\ng-v5\discover-packages.js:4:58)
    at <anonymous>
```

## Cannot read property of undefined in discover-packages.js

_A bug: Cannot read property of undefined in discover-packages.js
error getting on running npm-run pcakagr

_A feature: please describe your use case and motivation_

#### How To Reproduce

_A bug: please include instructions how to reproduce. Issues without reproduction are likely to receive no feedback._

_Can you reproduce the error in the integration tests in ng-packagr?_
_If possible, take a look at the [`integration/samples`](./integration/samples) and try to break one of these builds!_

_Is the error you faced in an application importing the library_
_Try to break the Angular CLI app in [`integration/consumers/ng-cli`](./integration/consumers/ng-cli)!_

#### Version Information

```
$ node_modules/.bin/ng-packagr --version
ng-packagr: 5.0.1
@angular/*: 7.X
typescript: 3.2.2
rxjs: 6.3.3
node:8.12
npm/yarn: 1.12.3
```

_Please include all version numbers that might be relevant, e.g. third-party libraries_

alan-agius4:
This seems like a bug that we should handle more gracefully. Do you have invalid `package.json`

PR #1256, opened 2019-04-01
fix: error out gracefully when `package.json` is not found

alan-agius4:
Closes: #1255


SaranVairam:
@alan-agius4  , what is the fix i need to give , for making an successful package

alan-agius4:
You most likely missing a `package.json` from a secondary entrypoint

SaranVairam:
@alan-agius4 ..  didnt get you , but any way ,  i will try ..Thanks

alan-agius4:
You most likely, have a missing `package.json` from a secondary entrypoint you are trying to import. This fix, will print out the path of the missing `package.json`

SaranVairam:
But the second entry point is same is defined in the entry as well console print out file

PR #1293, opened 2019-05-27
fix: don't write stacktraces when there are errors

alan-agius4:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
+++ b/src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
@@ -20,33 +20,31 @@ export const compileNgcTransform: Transform = transformFromPromise(async graph =
   const { basePath, cssUrl, styleIncludePaths } = entryPoint.data.entryPoint;
   const stylesheetProcessor = new StylesheetProcessor(basePath, cssUrl, styleIncludePaths);
 
-  await Promise.all([
-    compileSourceFiles(
-      graph,
-      entryPoint,
-      tsConfig,
-      moduleResolutionCache,
-      stylesheetProcessor,
-      {
-        outDir: path.dirname(esm2015),
-        declaration: true,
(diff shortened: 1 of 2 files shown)


src/lib/ng-v5/packagr.ts
```diff
@@ -117,8 +117,9 @@ export class NgPackagr {
       buildTransformOperator,
       catchError(err => {
         // Report error and re-throw to subscribers
-        log.error(err);
-        throw err;
+        const msg = process.env.DEBUG || !(err instanceof Error) ? err : err.message;
+        log.error(msg);
+        throw throwError(err);
```
JoostK: Was this meant to throw the `Observable` returned from `throwError`? I would expect `return` here.

src/lib/ng-v5/entry-point/ts/compile-ngc.transform.ts
```diff
@@ -20,33 +20,31 @@ export const compileNgcTransform: Transform = transformFromPromise(async graph =
...
-      moduleResolutionCache,
-      stylesheetProcessor,
-      {
-        outDir: path.dirname(esm2015),
-        declaration: true,
-        target: ts.ScriptTarget.ES2015
-      },
-      path.dirname(declarations)
-    ),
+  await compileSourceFiles(
```
JoostK: Can't these run concurrently anymore?

alan-agius4: The reason being if they run concurrently error messages are twice. That said, concurrently in most cases is slower since the IO ops are still a limiting factor.

src/lib/ng-v5/packagr.ts
alan-agius4: Opps good catch

PR #1346, opened 2019-07-24
feat: add support for Ivy libraries 

alan-agius4:
With this change we add support to generate Ivy libraries, we add NGCC in the module resolution pipeline to be able to transform VE libraries to NGTSC compatible code.

It is important to highlight that It is not recommended to publish Ivy libraries to NPM repositories.


alan-agius4:
JFYI @filipesilva 

@JoostK mind having a look? Since at this point, you are the most knowledge of ng-packagr after myself.

src/lib/ng-v5/init/init-tsconfig.transform.ts
```diff
@@ -1,12 +1,19 @@
...
+import { msg } from '../../util/log';
 
 export const initTsConfigTransformFactory = (defaultTsConfig: ParsedConfiguration): Transform =>
   transformFromPromise(async graph => {
     // Initialize tsconfig for each entry point
     const entryPoints = graph.filter(isEntryPoint) as EntryPointNode[];
     initializeTsConfig(defaultTsConfig, entryPoints);
+
+    if (defaultTsConfig.options.enableIvy) {
+      msg(chalk.yellow('\n** It is not recommended to publish Ivy libraries to NPM repositories **'));
```
JoostK: I would recommend we provide some more context here, and maybe use stronger wording than "not recommended". I think it's important to communicate what the consequences of publishing would be for the library. Maybe just link to an MD file in the repo with additional info, similar to schematics documentation.

src/lib/ngc/create-emit-callback.ts
```diff
@@ -1,87 +0,0 @@
...
- */
-
-// @link https://github.com/angular/angular/blob/24bf3e2a251634811096b939e61d63297934579e/packages/compiler-cli/src/transformers/util.ts#L14
-const GENERATED_FILES = /(.*?)\.(ngfactory|shim\.ngstyle|ngstyle|ngsummary)\.(js|d\.ts|ts)$/;
-
-import * as tsickle from 'tsickle';
-import * as api from '@angular/compiler-cli/src/transformers/api';
-
-// @link https://github.com/angular/angular/blob/83d207d/packages/compiler-cli/src/main.ts#L42-L84
-export function createEmitCallback(options: api.CompilerOptions): api.TsEmitCallback | undefined {
```
JoostK: Is this removing the ability to run the tsickle transforms? If so, why are we removing it?

src/lib/ngc/ngcc-processor.ts
```diff
@@ -0,0 +1,114 @@
+import { Logger, PathMappings } from '@angular/compiler-cli/ngcc';
+import { existsSync } from 'fs';
+import * as path from 'path';
+import * as ts from 'typescript';
+import * as log from '../util/log';
+import { EntryPointNode, ngUrl } from '../ng-v5/nodes';
+
+// Transform a package and it's typings when NGTSC is resolving a module.
```
JoostK:
```suggestion
// Transform a package and its typings when NGTSC is resolving a module.
```

src/lib/ngc/ngcc-processor.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  processModule(moduleName: string, resolvedModule: ts.ResolvedModule | ts.ResolvedTypeReferenceDirective): void {
+    const resolvedFileName = resolvedModule.resolvedFileName;
+    if (
+      !resolvedFileName ||
+      moduleName.startsWith('.') ||
+      this._processedModules.has(moduleName) ||
+      this._entryPointsUrl.includes(ngUrl(moduleName))
+    ) {
+      // Skip when module is unknown, relative or NGCC compiler is not found or already processed.
```
JoostK: This comment looks inaccurate. There's no check for `... NGCC compiler is not found ...`

<sup>It's also inaccurate in `@ngtools/webpack` ;)</sup>

src/lib/ngc/ngcc-processor.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      return;
+    }
+
+    this._ngcc.process({
+      basePath: this._nodeModulesDirectory,
+      targetEntryPointPath: path.dirname(packageJsonPath),
+      compileAllFormats: false,
+      propertiesToConsider: ['fesm2015', 'fesm5', 'es2015', 'esm2015', 'esm5', 'main', 'module'],
+      createNewEntryPointFormats: true,
```
JoostK: If we have this, ngcc generates its output in a separate subdirectory `__ivy_ngcc__` and the `package.json` file is updated with new entries for properties like `main`, to become suffixed with `_ivy_ngcc` targeting the `__ivy_ngcc__` subdirectory. I don't know how this is dealt with exactly, but I believe the CLI is responsible for switching over to these properties somehow, instead of the original. How is this dealt with in ng-packagr?

alan-agius4: In ng-packagr we only care about the typings for type checking, these are not placed under `__ivy_ngcc`.

Unfortunately ngcc doesn’t offer a way to process typings yet. That being said that is not a big deal. As the other module formats will probable be needed when running unit tests.

src/lib/ngc/ngcc-processor.ts
alan-agius4: Old comment 😜

src/lib/ng-v5/init/init-tsconfig.transform.ts
alan-agius4: Probably make sense to have something on AIO, eventually.

src/lib/ngc/ngcc-processor.ts
JoostK: Ah, that makes total sense. I see now, the `.d.ts` files are the new `metadata.json` :-)

src/lib/ngc/create-emit-callback.ts
alan-agius4: You are right.

alan-agius4:
@JoostK, PTAL

JoostK:
LGTM!

amitport:
so how do you "publish Ivy libraries to NPM repositories"? 
should just stick to angular8 / non-Ivy?

alan-agius4:
@amitport, you should not publish Ivy libraries to the NPM. In Angular version 9, you should still publish your library in VE mode, which will be converted to Ivy compatible code at an application level with NGCC.

This is mainly due the underlying instruction set might change in minor versions of Angular 9.

PR #1368, opened 2019-08-07
feat: show error when trying to publish ivy packages

alan-agius4:
Publishing of Ivy packages should not be allowed, as there Ivy packages are not backward compatible with VE applications.

With this change we use the npm/yarn prepublishOnly hook to display and error and abort the process with a non zero error code when a user tries to publish an Ivy version of the package.

More info: https://docs.npmjs.com/misc/scripts

### DIFF
--- a/integration/samples/apf/specs/package.ts
+++ b/integration/samples/apf/specs/package.ts
@@ -11,6 +11,10 @@ describe(`@sample/apf`, () => {
       expect(PACKAGE).to.be.ok;
     });
 
+    it(`should not have 'prepublishOnly' script`, () => {
+      expect(PACKAGE.scripts && PACKAGE.scripts.prepublishOnly).to.be.undefined;
+    });
+
     it(`should have 'tslib' as a dependency`, () => {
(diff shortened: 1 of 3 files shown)


mohammedzamakhan:
@alan-agius4 I am trying to create a v2 of one of my Angular Modules, and I am using v9.0.0-beta.3 which is compiles using Ivy by default. I also intent my module to be used only by v9.0.0 and up. 

I am getting the error reporting using this PR when I try to build. What should I be doing?

alan-agius4:
What version of ng-packagr are you using? And what is the error?

mohammedzamakhan:
v5.5.0

alan-agius4:
You need to opt-out from by setting `enableIvy` to `false` under `AngularCompilerOptions` In your tsconfig.

JoostK:
@alan-agius4 Could we configure ng-packagr to do that by default?

alan-agius4:
It is the default unless a custom tsconfig is used.

When a custom tsconfig is used the option is not set nor overridden.

Issue #1318, opened 2019-06-10
Configure the preferred main bundle


matheo:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

I'm building a library with `@nrwl/nx@8.1.0` and `ng-packagr@^5.2.0` for Firebase Functions. I've been deploying them successfully, but now that I'm trying to serve the built library with `firebase-tools` `emulate:start` I'm having an issue when it tries to compile the bundle:
```
⚠  TypeError: firebaseAdmin.hasOwnProperty is not a function
    at dist/functions/bundles/my-functions.umd.js:10:52
```
it's thrown at this line:
```
firebaseAdmin = firebaseAdmin && firebaseAdmin.hasOwnProperty('default') ? firebaseAdmin['default'] : firebaseAdmin;
```
and it's because `firebase-admin@8.0.0` exports a `FirebaseNamespace` esModule like this:
```
FirebaseNamespace {
>    __esModule: true,
>    credential:
>     { cert: [Function: cert],
>       refreshToken: [Function: refreshToken],
>       applicationDefault: [Function: applicationDefault] },
>    SDK_VERSION: '8.0.0',
>    Promise: [Function: Promise],
>    INTERNAL:
>     FirebaseNamespaceInternals {
>       firebase_:
>        FirebaseNamespace {
>          __esModule: true,
>          credential: [Object],
>          SDK_VERSION: '8.0.0',
>          Promise: [Function: Promise],
>          INTERNAL: [Circular],
>          default: [Circular] },
>       serviceFactories: {},
>       apps_: {},
>       appHooks_: {} },
>    default:
>     FirebaseNamespace {
>       __esModule: true,
>       credential:
>        { cert: [Function: cert],
>          refreshToken: [Function: refreshToken],
>          applicationDefault: [Function: applicationDefault] },
>       SDK_VERSION: '8.0.0',
>       Promise: [Function: Promise],
>       INTERNAL:
>        FirebaseNamespaceInternals {
>          firebase_: [Circular],
>          serviceFactories: {},
>          apps_: {},
>          appHooks_: {} },
>       default: [Circular] } }
```

#### Expected Behaviour

Modifying manually the `package.json` to use another bundle (like `esm5`) as the `main` package, it does run with no problems.

Only if we could have a new config option for the `ng-packagr.json` to pick the preferred bundle to be the main one in the resulting package.json, this issue could be avoided.

#### Version Information

```
[...]

PR #1372, opened 2019-08-10
feat: adds an option to choose library entry module type

jsaguet:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Description

This PR introduces a new option in ng-packagr.json to let the user decide which entry module type is preferred for the built library.
For example, it allows users to set esm5 module (instead of the default fesm5) in the built library's package.json "module" property which makes the library treeshakable when used somewhere else.

### How to use?
Simply add a "module" option in the "lib" property of ng-package.json file
Example ng-package.json:
```
{
    "$schema": "../../node_modules/ng-packagr/ng-package.schema.json",
    "dest": "../../dist/my-lib",
    "lib": {
      "entryFile": "src/public-api.ts",
      "module": "esm5"
    }
}
```

The possible values for the module property are "fesm5", "fesm2015", "esm5", "esm2015" as mentionned in the ng-package schema.

### Related issue
This PR can help for this issue: #1318 

```
[ ] Yes
[x] No
```

This PR does not introduce any breaking change since the "module" property is optional and defaults to fesm5 as it is currently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Create one `package.json` per npm package, run _ng-packagr_ for each!
 - [Embed Assets in CSS](docs/embed-assets-css.md)
 - [Managing Dependencies](docs/dependencies.md)
 - [Change the Entry File of a Library](docs/entry-file.md)
+- [Change the Library's Module type](docs/module-type.md)
 - [Change Configuration Locations](docs/configuration-locations.md)
 - [Override tsconfig](docs/override-tsconfig.md)
 - [Add Style Include Paths](docs/style-include-paths.md)
--- a/docs/module-type.md
+++ b/docs/module-type.md
@@ -1,0 +1,45 @@
+# Define Library's Module Type
+
+## Why?
+
+By default, the generated library's package.json `module` property points to the fesm5 bundle which is sometimes not treeshakable.
+
+You might want the `module` property of your library's package.json to point to the esm5 bundle for example.
+
+The same logic can be applied for `es2015` property if you want to point to esm2015 module instead of the default fesm2015.
+
+## How?
+
+### ES5 modules
+You can easily achieve this by adding a `module` property in the ng-package.json under lib:
+``` json
+{
+    "$schema": "../../node_modules/ng-packagr/ng-package.schema.json",
+    "dest": "../../dist/my-lib",
+    "lib": {
+      "entryFile": "src/public-api.ts",
+      "module": "esm5"
+    }
+}
+```
+
+Possible values for the module property are: 
+- fesm5 (default)
+- esm5
+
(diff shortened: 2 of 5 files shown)


alan-agius4:
Hi @jsaguet, thanks for the PR.

This really opens up quite a number of issues;

- While in some cases changing the module from FESM5 to ESM5 will improve threeshaking in some other cases this will have the opposite effect. see https://github.com/angular/angular-cli/pull/13650

- changing `module` to point to ES2015 or FESM2015 is definitely incorrect as `module` it is expected to have ES5 code. Also in the later versions of the CLI the `es2015` field  is preferred over the module fielding when having differential loading enabled.

- using ESM instead of FESM, will most likely cause slower build times because more modules will need to be resolved. (This needs to be confirmed as I am not sure if it’s still a problem for webpack)

- this will no longer conform with the APF specs.

jsaguet:
Hi @alan-agius4,
Thanks for your reply.

First of all, indeed, pointing module to esm2015 or fesm2015 is not right. I could implement the same option logic for es2015.

From what I've read in APF spec on the "[Note about the defaults in package.json](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview#heading=h.3kkc9d322g9f)", the goal is to point to esm files in the future. So it might be nice to give the option to the user to define esm modules directly if they have the need to do so.

As I mentioned in the PR, the default behavior would still point to fesm module so I guess this would still be compliant with APF spec, wouldn't it?

This is just about giving the choice to the user to customize the end package without having to modify it manually. I've had to do that personally and #1318 makes me think I'm not the only one.

Do you think this feature could be added?

Adrael:
Hi @alan-agius4,
First of all, thank you for the work you're doing for the community, amazing job!

I am facing the same issue as stated by @jsaguet.

While I agree with all the issues you listed, I have to agree with @jsaguet too - the default behaviour will not change and therefore none of the said issues should happen.
Developers will have the opportunity to document themselves and take the newly introduced `module` option for a spin.
Great power always come with great responsibility!

This PR would actually save me from doing some rather ugly changes on the generated package.json to properly tree-shake my apps.

Thanks for your time guys!

jsaguet:
Hi @alan-agius4, do you have any news regarding the integration of this feature in ng-packagr?
Thanks!

alan-agius4:
Thanks for the contribution. I think this will not get considering the potential upcoming changes in APF version 10.

I don’t think it is wise to add feature at this stage and remove it in a couple of months.

alan-agius4:
Closing as per above comment.

Issue #1395, opened 2019-09-03
Feature Request: Add option to remove fields from publish-ready package.json


elliotleelewis:
## Type of Issue

```
[ ] Bug Report
[*] Feature Request
```

## Description

It would be nice to be able to remove sections from the package.json after its been copied into the published output. For example, whilst under development, we add in our tooling as `devDependencies` in our `package.json` file, as well as things such as prettier/stylelint configs, which sometimes can cause issues, and are not needed in the final package.json.

#### Expected Behaviour

I propose that we add a `blacklistedPackageSections` field to the configuration file, so that we can specify properties to be removed when the file is built and copied over into the output directory. The settings file would then read as:

```
{
  "$schema": "./node_modules/ng-packagr/package.schema.json",
  "ngPackage": {
    "blacklistedPackageSections": [
      "devDependencies",
      "browserslist",
      "prettier",
      "stylelint"
    ]
  }
}
``` 

We could even remove the need for the `keepLifecycleScripts` config option too, by making the default for `blacklistedPackageSections` the following:

```
"blacklistedPackageSections": [
  "scripts"
]
```

What do you guys think? 😃

alan-agius4:
Kindly see: https://github.com/ng-packagr/ng-packagr/pull/1397#issuecomment-547820364

PR #1397, opened 2019-09-04
feat: added ability to blacklist package sections for removal in build

elliotleelewis:
## I'm submitting a...

```
[ ] Bug Fix
[*] Feature
```

## Checklist

- [*] Commit Messages follow the [Conventional Commits](https://conventionalcommits.org/) pattern
- [*] Tests for the changes have been added

## Description

Adds the feature requested in: #1395 

Adds the ability to blacklist sections in your package.json so that they're removed at build time. To use, add the `blacklistedPackageSections` value to your config object, and then include any sections from your package.json that you want to remove.

```
[ ] Yes
[*] No
```

### DIFF
--- a/integration/samples/blacklist-package-sections/package.json
+++ b/integration/samples/blacklist-package-sections/package.json
@@ -1,0 +1,28 @@
+{
+  "$schema": "../../../src/package.schema.json",
+  "name": "@sample/blacklist-package-sections",
+  "description": "A sample library with the blacklistPackageSections setting set",
+  "version": "1.0.0-pre.0",
+  "private": true,
+  "repository": "https://github.com/ng-packagr/ng-packagr.git",
+  "ngPackage": {
+    "lib": {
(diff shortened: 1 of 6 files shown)


alan-agius4:
Hi @elliotleelewisl, sorry for getting back to you so late.

I like the idea of removing certain sections from the `package.json`, however for now, I prefer to do so without exposing it via the public api.

I did implement something similar which is "private".
https://github.com/ng-packagr/ng-packagr/blob/57cc4d1caaca52507d19cbed470061cc72aa9ef1/src/lib/ng-v5/entry-point/write-package.transform.ts#L153-L161

elliotleelewis:
Hey @alan-agius4 !

No worries, the ones you added in "privately" fix the issues that we were having when using `ng-packagr`, so unless there is demand for this to be exposed, then this is good enough for now! Thanks a lot! 😄 

Is this going to be in the next major release then?

Issue #1400, opened 2019-09-13
Downlevel decorators to static fields without tsickle


alexeagle:
tsickle is intended to adapt TypeScript output to be valid Closure Compiler input.
One feature required is to downlevel decorators into static fields. We do it in a way that doesn't introduce Temporal Dead Zones.

Angular applications and libraries also downlevel decorators, so it was convenient to depend on the tsickle package for this. However it causes a bunch of support burden on tsickle, eg.
https://github.com/angular/tsickle/pull/922
where a language feature isn't used with closure compiler, so tsickle team doesn't plan to fix these.

For Angular apps, we fixed this by porting the downleveling code from tsickle to the webpack loader in
https://github.com/angular/angular-cli/pull/14473

We need to do something similar for ng-packagr so that Angular libraries can also be built without a tsickle dependency.

alexeagle:
Reading that angular-cli PR thread, it seems that https://github.com/angular/angular/issues/30586 is the root cause - it would be better not to require the downleveling at all.

PR #1401, opened 2019-09-13
fix: add downlevelConstructorParameters transformer

alan-agius4:
With this change tsickle is no longer needed to be enabled by default. Previously it was needed to TDZ that was caused by forwardRef.

Closes: #1400


alan-agius4:
From the CLI side, we need to add a migration to remove tsickle and turn off annoateForClosureCompiler.

PR #1406, opened 2019-09-25
fix: support the ng-package.json in secondary entry points

klemenoslaj:
Currently secondary entry points are breaking `no-implicit-dependencies` TSLint rule, because the
rule is checking for package.json files, which in secondary entry points are anyway just containing
`ngPackage` configuration. Adding the support for ng-package.json removes the need for package.json,
thus TSLint will look correctly in parent package.json file and act accordingly.

fix 1391

## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Fixes: [1391](https://github.com/ng-packagr/ng-packagr/issues/1391)

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples/secondary/feature-d/feature-d.ts
+++ b/integration/samples/secondary/feature-d/feature-d.ts
@@ -1,0 +1,5 @@
+import { FEATURE_A } from '@sample/secondary/feature-a';
+import { FEATURE_B } from '@sample/secondary/feature-b';
+import { FEATURE_C } from '@sample/secondary/feature-c';
+
+export const FEATURE_D = `Feature D: ${FEATURE_A} ${FEATURE_B} ${FEATURE_C}`;
--- a/src/lib/ng-v5/discover-packages.ts
+++ b/src/lib/ng-v5/discover-packages.ts
@@ -32,21 +32,22 @@ function formatSchemaValidationErrors(errors: ajv.ErrorObject[]): string {
  * Resolves a user's package by testing for 'package.json', 'ng-package.json', or 'ng-package.js'.
  *
  * @param folderPathOrFilePath A path pointing either to a file or a directory
+ * @param isSecondary A boolean determining if this is a secondary package
  * @return The user's package
  */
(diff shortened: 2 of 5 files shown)


klemenoslaj:
I as well wanted to add an addition to the documentation. Would it be fine like the example bellow?

`docs/secondary-entrypoints.md`
```markdown
### Alternative to `package.json`

Alternatively, you could create `ng-package.json` instead of `package.json`.
This is particularly useful in conjunction with `no-implicit-dependencies` TSLint rule, which will complain if `package.json` does not contain the dependencies used in the secondary entry point, which is misleading since all the dependencies should be mentioned in the primary `package.json`.
```

src/lib/ng-v5/discover-packages.ts
```diff
@@ -32,21 +32,22 @@ function formatSchemaValidationErrors(errors: ajv.ErrorObject[]): string {
  * Resolves a user's package by testing for 'package.json', 'ng-package.json', or 'ng-package.js'.
  *
  * @param folderPathOrFilePath A path pointing either to a file or a directory
+ * @param isSecondary A boolean determining if this is a secundary package
```
alan-agius4: secundary -> secondary

integration/samples/secondary/feature-d/feature-d.ts
```diff
@@ -0,0 +1,5 @@
+import { FEATURE_A } from '@sample/secondary/feature-a';
```
alan-agius4: Can we add a spec and verify that the generate `package.json` exists and contains all the expected fields.

Such as the correct `name` and other APF fields?

klemenoslaj:
@alan-agius4, how do you prefer in this projects for review adjustments to be applied? With additional commit or fixing the current commit?

If we're squashing it anyway makes no difference, just wanna be sure.

alan-agius4:
@klemenoslaj, doesn't make a difference :)

src/lib/ng-v5/discover-packages.ts
klemenoslaj: 🤦‍♂️

PR #1410, opened 2019-09-27
fix: disable internal `emitDecoratorMetadata`

alan-agius4:
t the moment, we are always setting `emitDecoratorMetadata`, this causes a problem in ES2015 when using `forwardRef`.

With https://github.com/ng-packagr/ng-packagr/pull/1401 `emitDecoratorMetadata` is no longer needed unless using `tsickle`.

`createEmitCallback` has been updated for this reason and also their is a PR for compiler-cli to do the same https://github.com/angular/angular/pull/32880

### DIFF
--- a/integration/consumers/ng-cli/package.json
+++ b/integration/consumers/ng-cli/package.json
@@ -27,8 +27,8 @@
     "zone.js": "^0.10.0"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


alan-agius4:
The downside is that there might be libraries which break during runtime due to TDZ.

Issue #1092, opened 2018-09-20
Suggest: add copy files options in schema


cipchk:
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Copy `README.md` file by default. But when have multiple language versions, like this:

```
README.md
README.zh-CN.md
```

I need to manually copy other language versions. 

Sometimes, project need dependencies original styles files (like `less`) in release package.

StudioSpindle:
Possible duplicate of #857

Splaktar:
This doesn't seem like a duplicate as the previous was reporting that the feature didn't work. But this one is suggesting that the feature should exist.

If we want SCSS support as well, should another feature request be opened or is it appropriate to discuss that here?

linnenschmidt:
Hi there, earlier this week I realized that there is no copy functionality for library assets.

Therefore I have extended the @angular-devkit/build-ng-packagr:build and created an own one with asset copy functionality.

Here you will find how to use it: https://github.com/linnenschmidt/build-ng-packagr#how-to-use-it

```bash
npm install @linnenschmidt/build-ng-packagr --save-dev
```
Please give it a try. Maybe this approach is enough for you.

shlomiassaf:
Lol, I have something similar as well, just more broad...

I have also built a replacement for `@angular-devkit/build-ng-packagr:build` called [ng-cli-packagr-tasks](https://github.com/shlomiassaf/ng-cli-packagr-tasks) that provide hooks to customise `ng-packagr` at various points in the build process.

One of the examples I show is how to [copy assests](https://github.com/shlomiassaf/ng-cli-packagr-tasks/blob/master/examples/copy-files.ts)

It's quite simple, but it will be in code because it supports a wide range of things (compile css, filter builds, etc..)

ferried:
For example. I want to publish my less. But the `ng-package.json`  has no option can do this.
It look's like copy files options: copy `*.less` to the `dest` file.
I need this option

redplane:
Any updates about this ? It is very time wasting when I have to copy all the asset files to dist folder.

Splaktar:
@redplane you can setup NPM tasks to do this in order to save time. Here's an [example](https://github.com/DevIntent/dev/blob/master/package.json#L16-L17).

grey280:
[...]

PR #1465, opened 2019-12-04
fix: add `assets` option to schema

alan-agius4:
This behaves the same way `srcs` of the `npm_package` bazel rule which allows copying of files to the distributable folder.

```js
  "assets": [
    'CHANGELOG.md',
    "./syles/**/*.scss"
  ],
  "lib": {
  }
```

Closes: https://github.com/angular/angular-cli/issues/11071 and closes: #1092


arvindnama:
Hi  , 
Can we know which version of ng-packagr has these changes .. i am on 5.7.1  and i dont see the `assets` option in the shema ? 
Or is this not released yet?

alan-agius4:
It will be stable by end of December

Splaktar:
Thank you so much for adding this capability! Are there any plans to document this in this repo?

arvindnama:
@alan-agius4  :: thanks for the update .. 
I tried using the above `asset` feature to copy assets in my library to package bundle , i noticed that the current feature will limit us to 
1. having all assets under project src .. ( from looking at the code)
we are  expecting assets to be under :: 
   projects/<lib-name>/**/assets
<dist-path>/<lib-name>/**/assets

with this i would not be able to use the ng-packagr feature as is .. 

So i spent some time enhancing this .. I was looking at the AsstePattern syntax of anuglar-builder ( browser) .. https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/browser/schema.json#L393

```
export type AssetPattern =
  | {
      /**
       * The pattern to match.
       */
      glob: string;
      /**
       */
      input: string;
      /**
       */
      ignore?: string[];
      /**
       */
      output: string;
    }
  | string;
```

I wanted this feature badly for my projects so did a poc for the same ,in the forked  repo https://github.com/arvindnama/ng-packagr..

wanted to know if it ok to make into `master` and what would be the procedure to do the same

Splaktar:
@arvindnama please open a new issue to propose your changes. Then after that, if you have a POC in another branch, you could open up a draft PR.

alan-agius4:
@Splaktar, yeah, indeed there is a plan to document this. At this time, I have limited time to dedicate to document this feature but it's in my todo list. Anyways PR's welcome to document this.

jrdutton:
Is there likely to be a 5.7.2 release with this feature included, please? I have included ng-packagr@9.0.0-rc.7 but am getting a number of peer dependency errors.

PR #1468, opened 2019-12-06
fix: add option to disable the prepublishOnly ivy publish script

### LINKED ISSUE #1467: Add option to publish a library built with ivy
## Type of Issue

```
[x] Feature Request
```

## Description

I have a library that is only for ivy, it uses ivy APIs that are not available in previous versions of the framework.
When I tried to publish it, I get the message:
> info:    err >  ERROR: Trying to publish a package that has been compiled by Ivy. This is not allowed.
> Please delete and rebuild the package, without compiling with Ivy, before attempting to publish.

#### Expected Behaviour
I should be able to publish my library for ivy anyway. I understand that it should be opt-in only because we want to avoid errors, but there should be a better way to allow that than to force me to manually edit the package.json file every time I want to build/publish my lib.
Ideally there should be an option in `ng-package.json` to do that.

#### Version Information

```
ng-packagr:            5.7.1
@angular/compiler:     9.0.0-rc.5
rollup:                1.25.2
tsickle:               0.37.1
typescript:            3.6.4
```

cc @alan-agius4

alan-agius4:
Fixes #1467

### DIFF
--- a/src/lib/ng-package-format/package.ts
+++ b/src/lib/ng-package-format/package.ts
@@ -62,10 +62,15 @@ export class NgPackage {
   }
 
   public get keepLifecycleScripts(): boolean {
-    return this.primary.$get('keepLifecycleScripts') === true;
+    return this.primary.$get('keepLifecycleScripts');
   }
+
   public get assets(): string[] {
     return this.primary.$get('assets');
+  }
+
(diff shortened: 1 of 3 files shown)


alan-agius4:
@JoostK let's see what @ocombe say's

PR #1500, opened 2020-01-07
feat: enhance asset pattern

arvindnama:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Recent PR #1465
Introduced the feature to copy assets.
But the current format will limit us to :

    1. Having all assets under project src .. ( from looking at the code)
    
i.e. we are expecting assets to be under ::
   ``` projects/<lib-name>/**/assets```
   ```<dist-path>/<lib-name>/**/assets```

with this i would not be able to use this ng-packagr feature as is ..
Expected Behaviour

I was looking at the AsstePattern syntax of anuglar-builder ( browser) .. https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/browser/schema.json#L393

```
export type AssetPattern =
  | {
      /**
       * The pattern to match.
       */
      glob: string;
      /**
       */
      input: string;
      /**
       */
      ignore?: string[];
      /**
       */
      output: string;
    }
  | string;
```

```
[ ] Yes
[x] No
```


Splaktar:
Thank you for the contribution. Please rebase when you get a chance.

arvindnama:
@Splaktar Done .. but `appveyor` fails .. the same thing works on other ci's and also on local

Splaktar:
Yeah, I saw that. Unfortunately, I'm not sure what's going on with those appveyor failures.

arvindnama:
@Splaktar  any update on the PR ?

Splaktar:
It's up to @alan-agius4 to do the review at this point and he's very busy with getting Angular version 9 shipped. This enhancement isn't required for version 9.

BioPhoton:
As v9 is out, could we restart the discussion?
I also run into this problem and it would be a very flexible solution IMOH.

Thanks @arvindnama btw for all the time!!

Sirius-A:
Awesome PR :+1: . I would really like to use this as well.

I just saw that this feature is not documented. I think it makes sense to add an example in `docs/copy-assets.md`. Or just mention that it follow the [Angular rules](https://angular.io/guide/workspace-config#assets-configuration).

arvindnama:
@alan-agius4  would be able to review the code .. i would love it have this feature in ngpackagr with ng-9 support ..

clecrois:
I'd like to use this as well! 😄

pburgmer:
This would be really useful to me too

4dams:
Are there any updates on this PR?

Splaktar:
It looks like this needs a rebase. I'll ping Alan in Slack about taking a brief look at this.

4dams:
> It looks like this needs a rebase. I'll ping Alan in Slack about taking a brief look at this.

Awesome, thanks a lot! :)

Splaktar:
The feedback that I got about this feature is that it may

> need some changes to make the files watchable

Sirius-A:
Thanks for reporting back :smile:. 

This MR is about extending the assets interface. Having it watch files would be a follow up topic imo. WDYT?

alan-agius4:
> Having it watch files would be a follow up topic imo.

The current implementation does watch file li s, therefore watching files will still need continue to work for all patterns.

Sirius-A:
Ah I see, in this case it certainly should do so as well.
Thanks for the clarification.

omnibrain:
Any reason not to merge this (other than the conflict)?

alan-agius4:
This causes a regression where assets will no longer be watched.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -42,20 +42,43 @@ export const writePackageTransform: Transform = transformFromPromise(async graph
     );
   }
   if (ngPackage.assets.length && !ngEntryPoint.isSecondaryEntryPoint) {
-    const assets = ngPackage.assets.map(x => path.join(ngPackage.src, x));
+    const ingoreOptions = {
+      ignore: ['.gitkeep', '**/.DS_Store', '**/Thumbs.db', '**/node_modules/**', `${ngPackage.dest}/**`],
```
sergiubologa, line 46: Why ignore the node_modules? This would be really useful for me because I use another library that comes with its own assets that need to be delivered with my library.

pongells:
> This causes a regression where assets will no longer be watched.

would a simpler implementation prevent that issue? e.g. just being able to configure output location

alan-agius4:
That is no longer a problem with the recent changes. I am okay with these changes, if anyone wants to create a PR.

Closing as this PR cannot be merged as it's stale.

CobusKruger:
> That is no longer a problem with the recent changes. I am okay with these changes, if anyone wants to create a PR.
> 
> Closing as this PR cannot be merged as it's stale.

@alan-agius4 Do I understand correctly that if we take the existing commit, rebase, and create a new PR, and don't mess it up somehow, you would be happy to merge it?

PR #1502, opened 2020-01-08
feat: resolve custom module alias from ts-path

arvindnama:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Current implementation does not resolve custom modules from ts-paths.

i.e.
ts-config.json
```
 {
 "paths": {
      "@ts-path/api": ["./src/api/index"]
    }
}
```

angular.component.js ( component in the library)

```
import { Component, Input } from '@angular/core';
import { getSampleApi } from '@ts-path/api';
@Component({
  selector: 'ng-component',
  styleUrls: ['./angular.component.scss'],
})
export class AngularComponent {
  @Input() title = 'Angular';
  constructor(){
    console.log(getSampleApi());
  }
}
```
here @ts-path is a custom module alias , who paths is present in ts-config,
current ng-packagr build does not resolve these modules based on ts-config path.

## Expected Behaviour

To Resolve module alias from ts-path

```
[ ] Yes
[x] No
```


JoostK:
I am very much against adding support for this. Dealing with import paths in libraries is incredibly finicky already, but some ground rules go a long way in avoiding potential issues.

Imports within the same entry-point **have to be** relative, otherwise they would be considered to import from an external package/entry-point. On the other hand, imports into other entry-points **must never be** relative. If these rules are not followed, you run into all sorts of issues. Conceptually, this difference is essential to be able to look at an import and know where it is resolved from (i.e. whether it crosses an entry-point boundary or not), so magically transforming absolute import paths based on path mappings is highly undesirable if you ask me.

Another problem with absolute module paths like this is that it is more likely to introduce imports into the public API from within the entry-point itself. Although this can work, it is very likely to cause circular dependency issues almost immediately. I understand that this isn't necessarily an issue when using _deep_ absolute import paths, however I consider those a bad practice in and of themselves (again, because you can no longer identify entry-point boundaries).

All in, adding support for magic transformations only makes things more complicated, in my opinion.

alan-agius4:
Hi @arvindnama, firstly I'd like to thank you for your contribution.

On this topic I agree with @JoostK, and I don't think we should support this feature. Currently this behaviour is supported in certain applications due to plugins for bundlers, and I personally think that this feature is being abused.

IMH, TypeScript path mappings shouldn't be used to make imports cleaner, shorter, sexier or whatever adjective you want to use.  But rather used to resolve dependencies at design time that would otherwise not be automatically resolved, which is what it was designed for.

Implementing this feature will also be harder for a library author, if they want to move away from `ng-packagr` to something else, such as `Bazel` were such feature doesn't exists.

Thanks once again.

philly-vanilly:
But just for clarification: this was supported before, no? We are using a tsconfig like this for the build:

`{
    "extends": "./tsconfig.json",
    "compilerOptions": {
        "paths": {
            "@foo-app/*": ["dist/libs/*"]
        }
    }
}`

and it seems it is not working anymore, after upgrading from ng-packagr 5 to 9

Issue #1510, opened 2020-01-09
Incorrect detection of `potentialDependencies` in `9.0.0-rc.7`


CarstenLeue:
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

I am seeing this error during compilation:

```
ERROR: Entry point @acoustic-content-sdk/ng-edit-api which is required by @acoustic-content-sdk/ng-edit doesn't exists.
```

This problem has not existed in `9.0.0-rc.4`. 

Cause: 

In the file `node_modules\ng-packagr\lib\ng-package\entry-point\analyse-sources.transform.js` the `analyseEntryPoint` method populates the `potentialDependencies` record. 

It uses `moduleName.startsWith(packageNode.data.primary.moduleId)` to tell if the module is such a dependency.

This leads to false positives if the package name is a prefix (and not just a namespace) e.g. for this case:

```json
{
  "name": "@acoustic-content-sdk/ng-edit",
  ...,
  "dependencies": {
    "@acoustic-content-sdk/api": "9.0.0",
    "@acoustic-content-sdk/component-api": "9.0.0",
    "@acoustic-content-sdk/component-edit": "9.0.0",
    "@acoustic-content-sdk/edit-api": "9.0.0",
    "@acoustic-content-sdk/ng-api": "9.0.0",
    "@acoustic-content-sdk/ng-edit-api": "9.0.0",
    "@acoustic-content-sdk/ng-utils": "9.0.0",
    "@acoustic-content-sdk/utils": "9.0.0"
  },
  ...
}
```

The module name found as a dependency is `@acoustic-content-sdk/ng-edit-api` while the module name is `@acoustic-content-sdk/ng-edit`, a prefix of the dependency. 

I think that the test is supposed to find secondary entry points, but in this case `@acoustic-content-sdk/ng-edit-api` is not a secondary entry point but an external dependency.

The check should probably be:

```ts
if ((moduleName === packageNode.data.primary.moduleId) || moduleName.startsWith(packageNode.data.primary.moduleId + '/'))
```

#### Version Information

```
ng-packagr:            9.0.0-rc.7
@angular/compiler:     9.0.0-rc.8
rollup:                1.29.0
tsickle:               0.38.0
typescript:            3.7.4
```

alan-agius4:
good catch

JoostK:
@CarstenLeue nice find, thanks for all the testing and reporting!

CarstenLeue:
[...]

PR #1511, opened 2020-01-09
fix: incorrect detection of potential dependent entry-points

alan-agius4:
Closes: #1510

### DIFF
--- a/src/lib/ng-package/entry-point/analyse-sources.transform.ts
+++ b/src/lib/ng-package/entry-point/analyse-sources.transform.ts
@@ -74,7 +74,7 @@ async function analyseEntryPoint(
       }
 
       const moduleName = moduleSpecifier.text;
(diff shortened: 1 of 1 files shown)


src/lib/ng-package/entry-point/analyse-sources.transform.ts
```diff
@@ -74,7 +74,7 @@ async function analyseEntryPoint(
       }
 
       const moduleName = moduleSpecifier.text;
-      if (moduleName.startsWith(packageNode.data.primary.moduleId)) {
+      if (moduleName.startsWith(`${packageNode.data.primary.moduleId}/`)) {
```
JoostK: Should we test for `moduleName === packageNode.data.primary.moduleId` as suggested by @CarstenLeue?

alan-agius4: Sry just saw this, yeah definitely. I forgot that part.

PR #1512, opened 2020-01-09
fix: analyse should cater for module name being the primary entry point

alan-agius4:


CarstenLeue:
The build for this is still failing

alan-agius4:
It wasn't released yet.

PR #1711, opened 2020-08-03
fix: use unaltered scripts when `keepLifecycleScripts` is enabled

SchnWalter:
## I'm submitting a...

```
[x] Bug Fix ?!?
[ ] Feature
```

## Checklist

## Description

It is common for packages to have a `prepublish` or a `prepublishOnly` script, something like `scripts/prepublish.js` from this repository. This PR updates `writePackageJson` to not replace an existing `prepublishOnly` script when the `keepLifecycleScripts` flag is provided.

~~And I've also added a commit that replaced all Unicode non-breaking spaces (`U+00A0`) with actual spaces, since the NBSPs aren't really needed in those cases.~~ Move to another PR.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -136,26 +136,27 @@ async function writePackageJson(
     throw e;
   }
 
-  // Removes scripts from package.json after build
-  if (packageJson.scripts) {
-    if (pkg.keepLifecycleScripts !== true) {
-      log.info(`Removing scripts section in package.json as it's considered a potential security vulnerability.`);
-      delete packageJson.scripts;
-    } else {
-      log.warn(
-        `You enabled keepLifecycleScripts explicitly. The scripts section in package.json will be published to npm.`,
-      );
+  // Process package.json scripts.
+  if (pkg.keepLifecycleScripts) {
+    if (packageJson.scripts) {
+      log.warn(`You enabled keepLifecycleScripts explicitly. The scripts section in package.json will be published to npm.`);
     }
-  }
+  } else {
+    if (packageJson.scripts) {
+      log.info(`Removing existing scripts section in package.json as it's considered a potential security vulnerability.`);
+      delete packageJson.scripts;
+    }
 
-  if (isIvy && !entryPoint.isSecondaryEntryPoint) {
-    const scripts = packageJson.scripts || (packageJson.scripts = {});
-    scripts.prepublishOnly =
-      'node --eval "console.error(\'' +
-      'ERROR: Trying to publish a package that has been compiled by Ivy. This is not allowed.\\n' +
-      'Please delete and rebuild the package, without compiling with Ivy, before attempting to publish.\\n' +
-      '\')" ' +
-      '&& exit 1';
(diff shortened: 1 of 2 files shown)


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -136,26 +136,27 @@ async function writePackageJson(
...
-  if (isIvy && !entryPoint.isSecondaryEntryPoint) {
-    const scripts = packageJson.scripts || (packageJson.scripts = {});
-    scripts.prepublishOnly =
-      'node --eval "console.error(\'' +
-      'ERROR: Trying to publish a package that has been compiled by Ivy. This is not allowed.\\n' +
-      'Please delete and rebuild the package, without compiling with Ivy, before attempting to publish.\\n' +
-      '\')" ' +
-      '&& exit 1';
+    if (isIvy && !entryPoint.isSecondaryEntryPoint) {
+      packageJson.scripts = {
```
alan-agius4, line 151: Wouldn’t this remove all existing scripts?

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -136,26 +136,27 @@ async function writePackageJson(
     throw e;
   }
 
-  // Removes scripts from package.json after build
-  if (packageJson.scripts) {
-    if (pkg.keepLifecycleScripts !== true) {
-      log.info(`Removing scripts section in package.json as it's considered a potential security vulnerability.`);
+  // Process package.json scripts.
+  if (pkg.keepLifecycleScripts) {
+    log.warn(
```
alan-agius4, line 141: We shouldn’t show the warning unless scripts exists.

SchnWalter: You are right. Wrapped this in a condition.

src/lib/ng-package/entry-point/write-package.transform.ts
SchnWalter: The scripts have already been deleted at this point; The only difference is that this code never gets to run, because it's in the else branch of `if (pkg.keepLifecycleScripts) {}`, thus keeping the scripts intact.

But, now that I'm looking at this, the terminology "lifecycle script" is missleading, since, we are talking about all NPM scripts, not just the `npm_lifecycle_event` scripts. What do you think is the best aproach? Should this flag be deprecated in favor of a new one with a more generic name, like `keepPackageScripts`?

alan-agius4: I think deprecated `keepLifecycleScripts` in a favour of `keepPackageScripts` is a good idea.

I seems that when `keepLifecycleScripts` is false, we are no longer adding the `prepublishOnly` script. The `compiled by Ivy. This is not allowed ` should also be added when the library is compiled by Ivy.

SchnWalter: 👍 I'll look into this. We'll also need to mention this in the flag info.

SchnWalter:
Closing in favor of https://github.com/ng-packagr/ng-packagr/pull/1722

Issue #1674, opened 2020-06-22
Ability to reuse umdModuleId definitions in secondary entry points


karan-kang:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description

We are building a large library that has several secondary  & tertiary entry points. At present, we have to repeat umdModuleIds in each secondary/tertiary point, which defeats the principle of DRY and is error prone.

It would be helpful if ng-packagr can pick the base list of umdModuleIds from primary endpoint (or ancestor chain).

#### Version Information

```
ng-packagr:            4.7.1
@angular/compiler:     7.2.15
rollup:                0.67.4
tsickle:               0.34.3
typescript:            3.2.4
```

astectus:
As I understood you  try to have a `shared` entry point so you could  use it in multiple entry points ?

SchnWalter:
I will look into this over the weekend.

We're building multiple (private) libraries with different names from the same code-base, and it's a pain to have them maintained. I'll make sure that the entry point details will have a public version of this information and I will have secondary entry points look into the primary entry point.

SchnWalter:
Besides `lib.umdModuleIds`, what else should the entry points inherit from their parent?

1) `keepLifecycleScripts`?
2) `whitelistedNonPeerDependencies`?
2) properly resolved `lib.styleIncludePaths`?
3) `lib.cssUrl`?

P.S. With https://github.com/ng-packagr/ng-packagr/pull/1737, all this data will be available when it's needed, it will be trivial to make these changes.

flash-me:
@SchnWalter Will this happen by default? If yes, isn't this a possible breaking change? Will we get the option the not inherit from the parent?

cheers
flash ⚡

SchnWalter:
I don't see why this would be a breaking change, you can't really have 2 different globals with different code. But, will soon become irrelevant, UMD builds will be removed in the next release of the AFP (most probably sometime after the new linker will be merged - if it hasn't been merged already)

flash-me:
Possible case would be e.g. if you provide secondary entry points exactly because of that. The primary entry point uses lets say public umdModuleIds and you have a secondary entry point for special environments with different umds.
In that case, it would be nice, if the secondary umdModuleIds are not overwritten with those of the parents.

UMD becoming deprecated is aprreciated. So why changing here sth anyway? :P

PR #1737, opened 2020-08-27
refactor: use data from parent entry point

SchnWalter:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

I have refactored the `NgEntryPoint` to make the API more clear:

* replaced the `secondaryData` NgEntryPoint parameter with a reference to the parent entry point
  https://github.com/ng-packagr/ng-packagr/blob/7dcbdf04a35a72b548c267d0b89708a7809036b8/src/lib/ng-package/entry-point/entry-point.ts#L58-L59
* add `NgEntryPoint.libraryDestinationPath`

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ng-package/entry-point/entry-point.spec.ts
+++ b/src/lib/ng-package/entry-point/entry-point.spec.ts
@@ -4,8 +4,8 @@ import { NgPackageConfig } from '../../../ng-package.schema';
 import { NgEntryPoint } from './entry-point';
 
 describe(`NgEntryPoint`, () => {
-  const testDestinationPath = '/example/entry-point/dest'
-  const testSourcePath = '/example/entry-point/src'
+  const testSourcePath: string = resolve('/example/entry-point/src');
+  const testDestinationPath: string = resolve('/example/entry-point/dest');
   let primaryEntryPoint: NgEntryPoint;
   let secondaryEntryPoint: NgEntryPoint;
   let packageJson: Record<string, any>;
@@ -31,9 +31,10 @@ describe(`NgEntryPoint`, () => {
     expect(primaryEntryPoint.basePath).to.equal(testSourcePath);
 
     // Test readonly getters:
-    expect(primaryEntryPoint.entryFilePath).to.equal(resolve(testSourcePath, 'public-api.ts'));
+    expect(primaryEntryPoint.entryFilePath).to.equal(join(testSourcePath, 'public-api.ts'));
     expect(primaryEntryPoint.isSecondaryEntryPoint).to.be.false;
-    expect(primaryEntryPoint.destinationPath).to.equal(resolve(testDestinationPath));
+    expect(primaryEntryPoint.destinationPath).to.equal(testDestinationPath);
+    expect(primaryEntryPoint.libraryDestinationPath).to.equal(testDestinationPath);
     expect(Object.keys(primaryEntryPoint.destinationFiles).length).to.equal(6);
     expect(primaryEntryPoint.entryFile).to.equal('public-api.ts');
     expect(primaryEntryPoint.cssUrl).to.be.undefined;  // TODO: should default to 'inline'.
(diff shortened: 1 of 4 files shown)


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -132,7 +132,7 @@ async function writePackageJson(
   try {
     checkNonPeerDependencies(packageJson, 'dependencies', whitelist);
   } catch (e) {
-    await rimraf(entryPoint.destinationPath);
+    await rimraf(entryPoint.libraryDestinationPath);
```
SchnWalter, line 135: Interesting, I think that I've introduced a regression that the tests didn't catch, I'll look into this and write a test if that's the case.

SchnWalter: Here, the old code caught the error, removed entry point destination directory and then the error was re-thrown in order to cancel the build.

I think that this change is better, because the whole library destination directory is removed and you're not left with a partial build.

src/lib/ng-package/entry-point/entry-point.spec.ts
```diff
@@ -33,7 +32,8 @@ describe(`NgEntryPoint`, () => {
     // Test readonly getters:
     expect(primaryEntryPoint.entryFilePath).to.equal(resolve(testSourcePath, 'public-api.ts'));
     expect(primaryEntryPoint.isSecondaryEntryPoint).to.be.false;
-    expect(primaryEntryPoint.destinationPath).to.equal(resolve(testDestinationPath));
+    expect(primaryEntryPoint.entryPointDestinationPath).to.equal(resolve(testDestinationPath));
+    expect(primaryEntryPoint.libraryDestinationPath).to.equal(resolve(testDestinationPath));
```
SchnWalter, line 36: Is `NgEntryPoint.destinationPath` a public API? Does it need to go through a deprecation phase?

SchnWalter:
@alan-agius4, could you please review and merge PR?

With the changes from this PR, entry points now have a reference to the parent entry point and it will allow us to implement all sort of features that depend on this information, for example we could have a fix for #1674

alan-agius4:
Hi @SchnWalter,

I’ll try to remove in the coming days

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: I think deleting the entire library makes sense in this case.

src/lib/ng-package/entry-point/entry-point.spec.ts
alan-agius4: Strictly speaking the changes here are part of the public API. While this shouldn't effect anyone. I think It would be better to leave such changes for a major version. (v12), as we definitely shouldn't break anyone in a minor even though this API is not stable at all.

SchnWalter: Ok, I'll rename these so that we don't have a breaking change. And I'll add a TODO so that we have better names in an upcoming major version.

Thank!

alan-agius4: Would there still be a breaking change because of the change of behaviour?  Or, maybe I am not understanding what you mean by renaming.

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: Would be great if you can do the rename as a separate commit.

src/lib/ng-package/entry-point/entry-point.spec.ts
SchnWalter: I'll add back the old `NgEntryPoint.destinationPath`, mark it as deprecated and  I will make sure that it works the same way, after all, it's just a getter.

Thanks for the quick reply.

src/lib/ng-package/entry-point/write-package.transform.ts
SchnWalter: Moved to a separate PR: #1845

src/lib/ng-package/entry-point/entry-point.spec.ts
SchnWalter: `NgEntryPoint.destinationPath` is back and it's no longer deprecated. And I've also reduced the amount of changes made in this PR.

alan-agius4:
Closing as stale.

PR #1761, opened 2020-09-29
refactor: use TypeScript generics for the Traversable BuildGraph

SchnWalter:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

Improves type safety for the BuildGraph.

Besides the type changes, only one extra type guard was added, but it should have no effect at runtime:
```
      const foundNode = graph.get(pkgUri);
      if (!isPackage(foundNode)) {
        return graph;
      }

      const ngPkg: PackageNode = foundNode;
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/graph/build-graph.ts
+++ b/src/lib/graph/build-graph.ts
@@ -1,9 +1,21 @@
 import { Node } from './node';
 
+export type SimplePredicate<T = Node> = {
+  (value: T, index: number): boolean;
+  and?: (criteria: SimplePredicate<T>) => SimplePredicate<T>;
+};
+
+export type ComplexPredicate<T = Node, R extends T = T> =
+  | SimplePredicate<T>
+  | {
+      (value: T, index: number): value is R;
+      and?: (criteria: ComplexPredicate<T, R>) => ComplexPredicate<T, R>;
+    };
+
 export interface Traversable<T> {
-  filter(by: (value: T, index: number) => boolean): T[];
-  find(by: (value: T, index: number) => boolean): T | undefined;
-  some(by: (value: T, index: number) => boolean): boolean;
+  filter<R extends T = T>(by: ComplexPredicate<T, R>): R[];
+  find<R extends T = T>(by: ComplexPredicate<T, R>): R | undefined;
+  some(by: SimplePredicate<T>): boolean;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/lib/graph/build-graph.ts
```diff
@@ -47,20 +59,21 @@ export class BuildGraph implements Traversable<Node> {
...
-  public find(by: (value: Node, index: number) => boolean): Node | undefined {
-    return this.entries().find(by);
+  public find<T extends Node = Node>(by: ComplexPredicate<Node, T>): T | undefined {
+    return this.entries().find(by) as T;
   }
 
   public from(node: Node): Traversable<Node> {
-    return node;
+    // TODO: Remove or fix this method.
+    return (node as unknown) as Traversable<Node>;
```
SchnWalter, line 76: This method isn't used. Should we remove it?

alan-agius4: Yes please

src/lib/ng-package/package.transform.ts
```diff
@@ -202,7 +207,7 @@ const buildTransformFactory = (project: string, analyseSourcesTransform: Transfo
     scheduleEntryPoints(entryPointTransform),
     // Write npm package to dest folder
     writeNpmPackage(pkgUri),
-    tap(graph => {
+    tap((graph: BuildGraph) => {
```
alan-agius4, line 210: Are all these changes necessary? Shouldn't be these inferred?

alan-agius4 (commented):
Are all these changes necessary? Shouldn't these be inferred?

SchnWalter:
> Are all these changes necessary? Shouldn't these be inferred?

@alan-agius4, I've removed all the extra `switchMap((graph: BuildGraph)` changes. This MR now includes only the type casting changes.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -11,9 +11,9 @@ import { EntryPointNode, isEntryPointInProgress, isPackage, PackageNode, fileUrl
 import { Node } from '../../graph/node';
 
 export const writePackageTransform: Transform = transformFromPromise(async graph => {
-  const entryPoint = graph.find(isEntryPointInProgress()) as EntryPointNode;
+  const entryPoint: EntryPointNode = graph.find(isEntryPointInProgress());
   const ngEntryPoint: NgEntryPoint = entryPoint.data.entryPoint;
-  const ngPackageNode = graph.find(isPackage) as PackageNode;
+  const ngPackageNode: PackageNode = graph.find(isPackage);
```
SchnWalter, line 16: @alan-agius4, do you prefer using inferred types in all situations, like:
```
const ngPackageNode = graph.find(isPackage);
```
instead of
```
const ngPackageNode: PackageNode = graph.find(isPackage);
```
This will work now with any type casting, but I've added the types just to inform the user that now we have a different type. When dealing with type guards I usually add the type to the returned value. But let me know if you want to remove them; we should strive for consistency.

alan-agius4 (approved):
Thanks

PR #1764, opened 2020-10-10
Performance optimizations

JoostK:
See individual commits.

### DIFF
--- a/src/lib/ng-package/ngcc-cache.ts
+++ b/src/lib/ng-package/ngcc-cache.ts
@@ -1,7 +1,7 @@
 /**
- * Registers the paths to package.json files of libraries that have been processed by ngcc. This
- * cache is reused across all entry-points of a package, so module requests across the entry-points
- * can determine whether invoking ngcc is necessary.
+ * Registers the absolute specifiers of libraries that have been processed by ngcc. This cache is
+ * reused across all entry-points of a package, so module requests across the entry-points can
+ * determine whether invoking ngcc is necessary.
  *
  * The cost of invoking ngcc for an entry-point that has already been processed is limited due to
  * a fast path in ngcc, however even in this fast-path does ngcc scan the entry-point to determine
@@ -9,17 +9,17 @@
  * are processed in batches during which the `node_modules` directory is not mutated.
  */
 export class NgccProcessingCache {
-  private readonly processedPackageJsonPaths = new Set<string>();
+  private readonly processedModuleNames = new Set<string>();
 
-  hasProcessed(packageJsonPath: string): boolean {
-    return this.processedPackageJsonPaths.has(packageJsonPath);
+  hasProcessed(moduleName: string): boolean {
+    return this.processedModuleNames.has(moduleName);
   }
 
-  markProcessed(packageJsonPath: string): void {
-    this.processedPackageJsonPaths.add(packageJsonPath);
+  markProcessed(moduleName: string): void {
+    this.processedModuleNames.add(moduleName);
   }
 
   clear(): void {
-    this.processedPackageJsonPaths.clear();
+    this.processedModuleNames.clear();
   }
 }
--- a/src/lib/ng-package/ngcc-cache.ts
+++ b/src/lib/ng-package/ngcc-cache.ts
@@ -1,0 +1,25 @@
+/**
+ * Registers the paths to package.json files of libraries that have been processed by ngcc. This
+ * cache is reused across all entry-points of a package, so module requests across the entry-points
+ * can determine whether invoking ngcc is necessary.
+ *
+ * The cost of invoking ngcc for an entry-point that has already been processed is limited due to
+ * a fast path in ngcc, however even in this fast-path does ngcc scan the entry-point to determine
+ * if all dependencies have been processed. This cache allows to avoid that work, as entry-points
+ * are processed in batches during which the `node_modules` directory is not mutated.
+ */
+export class NgccProcessingCache {
+  private readonly processedPackageJsonPaths = new Set<string>();
+
+  hasProcessed(packageJsonPath: string): boolean {
+    return this.processedPackageJsonPaths.has(packageJsonPath);
+  }
+
+  markProcessed(packageJsonPath: string): void {
+    this.processedPackageJsonPaths.add(packageJsonPath);
+  }
+
+  clear(): void {
+    this.processedPackageJsonPaths.clear();
+  }
+}
--- a/src/lib/ng-package/package.transform.ts
+++ b/src/lib/ng-package/package.transform.ts
@@ -101,7 +101,11 @@ export const packageTransformFactory = (
       const ngPkg: PackageNode = foundNode;
       const entryPoints = [ngPkg.data.primary, ...ngPkg.data.secondaries].map(entryPoint => {
         const { destinationFiles, moduleId } = entryPoint;
-        const node = new EntryPointNode(ngUrl(moduleId), ngPkg.cache.sourcesFileCache);
+        const node = new EntryPointNode(
+          ngUrl(moduleId),
+          ngPkg.cache.sourcesFileCache,
+          ngPkg.cache.moduleResolutionCache,
+        );
         node.data = { entryPoint, destinationFiles };
         node.state = 'dirty';
         ngPkg.dependsOn(node);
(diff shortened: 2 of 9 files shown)


src/lib/ts/tsconfig.ts
```diff
@@ -32,6 +32,9 @@ export function readDefaultTsConfig(fileName?: string): ng.ParsedConfiguration {
     // these are required to set the appropriate EmitFlags
     flatModuleId: 'AUTOGENERATED',
     flatModuleOutFile: 'AUTOGENERATED',
+
+    // performance
+    skipLibCheck: true, // don't type-check .d.ts files for performance and memory reasons.
```
JoostK, line 37: There's also `skipDefaultLibCheck` which would continue to type-check d.ts files, but that one is deprecated in favor of `skipLibCheck` itself. See also microsoft/TypeScript#31417 for a relevant discussion on this flag.

src/lib/ng-package/package.transform.ts
```diff
@@ -134,9 +139,11 @@ const watchTransformFactory = (
       return createFileWatch(data.src, [data.dest]).pipe(
         tap(fileChange => {
           const { filePath, event } = fileChange;
-          const { sourcesFileCache } = cache;
+          const { sourcesFileCache, ngccProcessingCache } = cache;
           const cachedSourceFile = sourcesFileCache.get(filePath);
 
+          ngccProcessingCache.clear();
```
alan-agius4, line 145: Do we need to clear the cache on each file file? We don't watch node_modules therefore I assume that we shouldn't do this.

src/lib/ts/tsconfig.ts
alan-agius4: I think we should leave this one out.

src/lib/ng-package/ngcc-cache.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Registers the paths to package.json files of libraries that have been processed by ngcc. This
+ * cache is reused across all entry-points of a package, so module requests across the entry-points
+ * can determine whether invoking ngcc is necessary.
+ *
+ * The cost of invoking ngcc for an entry-point that has already been processed is limited due to
+ * a fast path in ngcc, however even in this fast-path does ngcc scan the entry-point to determine
+ * if all dependencies have been processed. This cache allows to avoid that work, as entry-points
+ * are processed in batches during which the `node_modules` directory is not mutated.
+ */
+export class NgccProcessingCache {
```
alan-agius4, line 11: Why not simiply use a `Set` directly?

src/lib/ngc/ngcc-processor.ts
```diff
@@ -36,7 +43,7 @@ export class NgccProcessor {
     }
 
     const packageJsonPath = this.tryResolvePackage(moduleName, resolvedFileName);
-    if (!packageJsonPath) {
+    if (!packageJsonPath || this.ngccProcessingCache.hasProcessed(packageJsonPath)) {
```
alan-agius4, line 46: Don't you think that we can unify ngccProcessingCache / processedModules? More or less they serve the same purpose.

alan-agius4 (commented):
Thanks for this @JoostK.

Couple of minor comments.

src/lib/ng-package/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+/** Lazily initialized ajv validator instance. */
+let ajvValidator: ajv.ValidateFunction | null = null;
+
+/**
+ * Validates the `ngPackageJson` value against the JSON schema using ajv. An error is thrown if
+ * schema errors are found.
+ *
+ * @param ngPackageJson The value to validate.
+ */
+export function validateNgPackageSchema(ngPackageJson: unknown): asserts ngPackageJson is NgPackageConfig {
```
alan-agius4, line 14: 👍

src/lib/ng-package/schema.ts
```diff
@@ -0,0 +1,67 @@
+import * as ajv from 'ajv';
+import { NgPackageConfig } from '../../ng-package.schema';
+import * as log from '../utils/log';
+
+/** Lazily initialized ajv validator instance. */
+let ajvValidator: ajv.ValidateFunction | null = null;
```
alan-agius4, line 6: Personal preference

```suggestion
let ajvValidator: ajv.ValidateFunction | undefined;
```

src/lib/ng-package/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+      return message + '.';
+    })
+    .join('\n');
+}
+
+/**
+ * Returns an initialized ajv validator for the ng-package JSON schema.
+ */
+function getNgPackageSchemaValidator(): ajv.ValidateFunction {
+  if (ajvValidator !== null) {
```
alan-agius4, line 41:
```suggestion
  if (ajvValidator) {
```

src/lib/ngc/ngcc-processor.ts
JoostK: The local `processedModules` still allows to avoid the `tryResolvePackage` call which does FS-based lookup, so it still has some benefit.

src/lib/ng-package/ngcc-cache.ts
JoostK: I like the documentation aspect of having the dedicated class (JS docblock, method names, etc.). It also provides type-safety with respect to other `Set` instances, so that's nice as well.

src/lib/ng-package/package.transform.ts
JoostK: I wanted to avoid a situation where `node_modules` was somehow affected between watch builds, as in that case the cache would incorrectly cause ngcc not to run. Even if we don't watch `node_modules` it's still possible for someone to change a file that is watched after something in `node_modules` has changed.

src/lib/ng-package/schema.ts
JoostK: Oh, interesting; I personally much prefer the explicit initializer and explicit non-null check.

alan-agius4: You can leave it :)

src/lib/ng-package/package.transform.ts
alan-agius4: I think this will not work because of the "local" processed files cache `processedModules`

JoostK: `NgccProcessor` is recreated in the transform pipeline, so it's not a persistent cache per entry-point.

src/lib/ngc/ngcc-processor.ts
JoostK: I see. I didn't do that because `tryResolvePackage` uses both `moduleName` and `resolvedFileName`, where the latter depends on the resolution context as determined by the entry-point. So caching based on `moduleName` only directly in the `NgccProcessingCache` might not be fully equivalent.

src/lib/ng-package/ngcc-cache.ts
alan-agius4: While you have a point with docs for methods which are not added here :P. 

I don't think it adds anything from a type-safety perspective, both the below are equity type-safe, unless I am missing something.
```ts
const ngccProcessingCache = new Set<string>();
const ngccProcessingCache = new NgccProcessingCache();
```

I don't have a very strong option about this and you can keep it as is. But I do think that in this case using a class doesn't add any value, especially for an internal implementation on top of that which is used in a single place.

Also, while in this case they are not in use, wrapping the Set with a Class has also some disadvantages where you don't expose all the methods and properties available for it. Example, if you need to use `.size` or `entries`, you'd need to implement another method like such https://github.com/ng-packagr/ng-packagr/blob/master/src/lib/file-system/file-cache.ts

JoostK: What I meant by type-safety is the fact that two `Set`s can easily be interchanged, which is not the case with two distinct implementations. I phrased the earlier comment quite poorly, hopefully this makes more sense. It's similar to the `AbsoluteFsPath` branded type that's used in the compiler; although it is a string it does offer identification/documentation and more type-safety compared to just using a string.

> While you have a point with docs for methods which are not added here :P.

😅 I was mainly referring to the docblock on the class itself.

I see what you're saying from an internal implementation perspective and the fact that this thing is used in so few places. I still like the insight in brings when just glancing over the code, and e.g. the ability to ask an IDE where things are marked as processed (which wouldn't be as easy if this were just `Set`).

src/lib/ngc/ngcc-processor.ts
JoostK: Simplified to just one cache per our discussion.

alan-agius4 (approved):
Thanks a lot of this!

Issue #1562, opened 2020-03-08
Lazy Loaded component & secondary entry point


GrandSchtroumpf:
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description
I've got a lib where the primary entry point has a map that lazy load components from the secondary entry point : 
```typescript
export const materialFormFactory = {
  text: () => import('material-form-factory/text').then(c => c.TextFormComponent),
  select: () => import('material-form-factory/select').then(c => c.SelectFormComponent),
  autocomplete: () => import('material-form-factory/autocomplete').then(c => c.FormAutocompleteComponent),
  entity: () => import('material-form-factory/entity').then(c => c.EntityComponent),
  list: () => import('material-form-factory/control-list').then(c => c.ControlListComponent)
} as const
```
When trying to build it I've got this error: 
```
Cannot find module 'material-form-factory/text'.
```

**Workaround** :
By adding an import to **one of the secondary entry point** it solves the issue : 
```typescript
// This is just here to solve build issue
import { TextFormComponent } from 'material-form-factory/text';

export const materialFormFactory = {
  text: () => import('material-form-factory/text').then(c => c.TextFormComponent),
  select: () => import('material-form-factory/select').then(c => c.SelectFormComponent),
  autocomplete: () => import('material-form-factory/autocomplete').then(c => c.FormAutocompleteComponent),
  entity: () => import('material-form-factory/entity').then(c => c.EntityComponent),
  list: () => import('material-form-factory/control-list').then(c => c.ControlListComponent)
} as const
```` 

#### How To Reproduce
Create an angular 9 project.
Add a library
Create several secondar entry points.
In the primary entry point (or any secondary entry point), create a map that load component lazily.

#### Expected Behaviour
I should be able to build without adding an import to one of the entry point.

#### Version Information
```
$ node_modules/.bin/ng-packagr --version
ng-packagr:  9.0.2
@angular/compiler: 9.0.0
rollup: 1.31.1
tsickle:               
typescript: 3.7.5
```

vmcbaptista:
This still happens in version 11

alan-agius4:
Hi lazy-loading is not meant to be used at a library level. And this approach would cause failures during runtime.

See: https://github.com/angular/angular/issues/31893#issuecomment-516615243 and https://github.com/angular/angular-cli/issues/6373

In future this might be possible when Ivy libraries are published. But until then this is not a viable approach with View Engine.

PR #1790, opened 2020-11-15
fix: consider dynamic imports while evaluating potential dependencies

vmcbaptista:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Pretends to #1562 .

I believe that the bug reported on #1562 happens because the dependencies analysis does not consider dynamic imports.

This PR is an attempt to also consider dynamic imports when evaluating dependencies.
Please consider this as a PoC. Probably the new parsing I'm introducing could be improved.
I'm open to any suggestions/improvements.

```
[ ] Yes
[ ] No
```
I'm uncertain about if this might cause unexpected breaking results for existing ng-packagr users.


alan-agius4:
Hi @vmcbaptista, lazy-loading is not meant to be used at a library level. And this approach would cause failures during runtime.

See: https://github.com/angular/angular/issues/31893#issuecomment-516615243 and https://github.com/angular/angular-cli/issues/6373

In future this might be possible when Ivy libraries are published. But until then this is not a viable approach with View Engine.

PR #1796, opened 2020-11-21
fix: correctly resolve sourceRoot

alan-agius4:
Closes https://github.com/angular/angular-cli/issues/19456


johncrim:
Thank you!!

PR #1838, opened 2021-01-20
feat: add NGCC async integration

alan-agius4:
With this change we run the async version of NGCC prior to module resolution.

Fixes: https://github.com/angular/angular-cli/issues/19817


MickL:
So fast, thank you!!

PR #1840, opened 2021-01-20
build: update build to remove copyfiles and cpx from dev-dependencies

alan-agius4:


src/lib/ng-package/package.transform.ts
```diff
@@ -237,7 +237,7 @@ const writeNpmPackage = (pkgUri: string): Transform =>
   pipe(
     switchMap(async graph => {
       const { data } = graph.get(pkgUri);
-      const files = [`${data.src}/LICENSE`, `${data.src}/README.md`, `${data.src}/CHANGELOG.md`];
+      const files = [`${data.src}/LICENSE`, `${data.src}/README.md`];
```
SchnWalter, line 240: With this change, the CHANGELOG.md files from libraries will no longer be included automatically, we should probably document this change, and while we're at it, maybe we could also look into removing the LICENSE entry from here and inform users that they need to add these files to their assets config.

> https://github.com/ng-packagr/ng-packagr/issues/1451

PR #1845, opened 2021-01-23
test: fix missing dependency build cleanup test

SchnWalter:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

~~Errors thrown in `writePackageJson()` for secondary entry point builds would cause the output directory of the library to contain a partial build of the library.~~

Fixes a broken test that was always passing.

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples/fail-missing-deps/package.json
+++ b/integration/samples/fail-missing-deps/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "non-peer-deps",
+  "name": "fail-missing-deps",
   "description": "https://github.com/ng-packagr/ng-packagr/pull/687",
   "dependencies": {
     "foobar": "1.0.0"
--- a/integration/samples/fail-missing-deps/README.md
+++ b/integration/samples/fail-missing-deps/README.md
@@ -1,3 +1,3 @@
-## failures
+## fail-missing-deps
 
 The build is expected to fail because of a missing dependency, causing the output directory to not exist.
--- a/integration/samples/fail-missing-deps/index.ts
+++ b/integration/samples/fail-missing-deps/index.ts
@@ -1,0 +1,1 @@
+export const HAS_MISSING_DEPENDENCY = 'should fail';
--- a/integration/samples/fail-missing-deps/specs/fail-missing-deps.ts
+++ b/integration/samples/fail-missing-deps/specs/fail-missing-deps.ts
@@ -2,8 +2,8 @@ import { expect } from 'chai';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 
-describe(`Failure Builds`, () => {
-  describe(`non-peer-deps`, () => {
+describe(`Missing primary entry point dependencies `, () => {
+  describe(`library build erorr`, () => {
     it(`should have no build output`, () => {
       const exists = fs.existsSync(path.resolve(__dirname, '../dist'));
       expect(exists).to.be.false;
--- a/integration/samples/failures/README.md
+++ b/integration/samples/failures/README.md
@@ -1,0 +1,3 @@
+## failures
+
+The build is expected to fail because of a missing dependency, causing the output directory to not exist.
--- a/integration/samples/failures/index.ts
+++ b/integration/samples/failures/index.ts
@@ -1,1 +1,0 @@
-export const PUBLISHING_NON_PEER_DEPENDENCIES = 'should fail';
--- a/integration/samples/failures/package.json
+++ b/integration/samples/failures/package.json
@@ -6,6 +6,7 @@
   },
   "ngPackage": {
     "dest": "./dist",
+    "deleteDestPath": true,
     "lib": {
       "entryFile": "index.ts"
     }
--- a/integration/samples/failures/specs/non-peer-deps.ts
+++ b/integration/samples/failures/specs/non-peer-deps.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 describe(`Failure Builds`, () => {
   describe(`non-peer-deps`, () => {
     it(`should have no build output`, () => {
-      const exists = fs.existsSync(path.resolve(__dirname, 'dist'));
+      const exists = fs.existsSync(path.resolve(__dirname, '../dist'));
       expect(exists).to.be.false;
     });
   });


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -185,7 +185,7 @@ async function writePackageJson(
   try {
     checkNonPeerDependencies(packageJson, 'dependencies', whitelist, spinner);
   } catch (e) {
-    await rimraf(entryPoint.destinationPath);
+    await rimraf(entryPoint.libraryDestinationPath);
```
SchnWalter, line 188: @alan-agius4, it turns out the output directory contain the most of the files for the secondary entry point, but not the directory with its package.json. This PR fixes the issue and adds tests.

Once this is merged, I'll continue with the #1737.

integration/samples.sh
```diff
@@ -27,7 +27,7 @@ do
         node dist/cli/main.js -p ${P_JSON}
     elif [ -f $P ]; then
         echo "Building sample ${P}..."
-        if [ $P='failures' ]; then
+        if [[ $P = *'fail-'* ]]; then
```
SchnWalter, line 30: @alan-agius4, this line used to have an issue, the old code was never using the else branch because:
```
$ P="abcd"

$ [ $P="failures" ] && echo "TRUE" || echo "FALSE"
TRUE

$ [ $P = "failures" ] && echo "TRUE" || echo "FALSE"
FALSE

$ [ $P = "abcd" ] && echo "TRUE" || echo "FALSE"
TRUE

$ [ $P = "bc" ] && echo "TRUE" || echo "FALSE"
FALSE

$ [[ $P == *"bc"* ]] && echo "TRUE" || echo "FALSE"
TRUE
```
The new code is bash specific, but since the script bang is `#!/bin/bash`, I assume that we can go with it. If we want to support more shells, we should change the bang to `#!/bin/sh` and use a POSIX compliant comparison here.

Also, the `$FIND_COMMAND` from the else branch is wrong since it uses glob expansion and not executing the bash command. And the `win32` branch might also have an issue, it doesn't look just into the samples directory. I'll take a look at this script in the future and I'll create a PR.

integration/samples/fail-missing-secondary-deps/secondary/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
alan-agius4, line 1: Can we remove this test?
This is an invalid entry-point. Dependencies cannot be defined in secondary entry-points.

SchnWalter: I've removed this test and added one with a secondary entry point that fails at:
```
import { exampleFooBarFactory } from '@angular/core';
```

The newly added test was failing, it turns out, that also in this case the destination was not cleaned.

I've tried looking for a single place to cleanup, mainly `entryPointTransformFactory()`, but I've ended up adding a cleanup inside all the transforms provided to `entryPointTransformFactory()`:
```
    compileTs,
    writeBundles,
    writePackage,
```

integration/samples.sh
SchnWalter: This change no longer exists, it has been fixed in #1846

integration/samples/fail-missing-secondary-deps/secondary/package.json
SchnWalter: It's too much of a hassle to cleanup on build errors and not on watch. I've kept the newly added tests that are failing, but I've disabled them.

I've recreated the branch, please check the new commits (they should not be squashed upon merging).

src/lib/ng-package/entry-point/entry-point.ts
```diff
@@ -61,45 +61,44 @@ export class NgEntryPoint {
...
   /** Absolute file path of the entry point's source code entry file. */
   public get entryFilePath(): string {
-    return path.resolve(this.basePath, this.entryFile);
+    return resolve(this.basePath, this.entryFile);
   }
 
   /** Whether or not the entrypoint is secondary */
   public get isSecondaryEntryPoint(): boolean {
-    return !!this.secondaryData;
+    return this.secondaryData !== undefined;
```
alan-agius4, line 69: Nit: in general I prefer the previous style.

src/lib/ng-package/entry-point/entry-point.transform.ts
```diff
@@ -41,7 +41,7 @@ export const entryPointTransformFactory = (
     //tap(() => log.info(`Building from sources for entry point`)),
 
     transformFromPromise(async (graph) => {
-      // Peek the first entry point from the graph
+      // Pick the first entry point from the graph
```
alan-agius4, line 44: `Peek` is the correct terms here.

src/lib/ng-package/entry-point/entry-point.ts
```diff
@@ -130,10 +129,22 @@ export class NgEntryPoint {
     return this.$get('lib.flatModuleFile') || this.flattenModuleId('-');
   }
 
+  /**
+   * Returns the relative source path for secondary entry points.
+   *
+   * TODO: Compare the primary entry point base path with the base path for this entry point and rename to "relativeBasePath".
+   */
+  private get secondaryDir(): string {
```
alan-agius4, line 137: Nit: Since this is only used once. I prefer to to create a method for it. A better cleaner way, would also to provide the directory name, when creating the entry-point.

https://github.com/ng-packagr/ng-packagr/blob/cd2f3d54973b3a68dede9e77953e60d9bb7cc3e2/src/lib/ng-package/discover-packages.ts#L125

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -132,7 +132,10 @@ export const writePackageTransform: Transform = transformFromPromise(async graph
  * flattened JavaScript bundles, type definitions, (...).
  *
  * @param entryPoint An entry point of an Angular package / library
```
alan-agius4, line 134: 👍

src/lib/ng-package/entry-point/entry-point.ts
SchnWalter: Converted to a method and moved all (two) methods to the bottom of the class.

As for the code that you've linked, I would keep this as it is, because it will be removed in https://github.com/ng-packagr/ng-packagr/pull/1737.

Thanks for the time!

src/lib/ng-package/entry-point/entry-point.ts
SchnWalter: Reverted.

src/lib/ng-package/entry-point/entry-point.transform.ts
SchnWalter: Reverted.

integration/samples/fail-missing-secondary-deps/secondary/package.json
SchnWalter: Resolving, the newly added disabled tests were removed, I'll create a followup PR when we also have a fix for the cleanup.

src/lib/ng-package/entry-point/entry-point.ts
SchnWalter: This change is no longer is this MR. Sorry for all the ping-pong and the amount of changes.

I've reduced this PR to the bare minimum, updated the descrdiption am moved some changes back to #1737, which now includes just 2 commits.

integration/samples/fail-missing-deps/package.json
```diff
@@ -1,11 +1,12 @@
 {
-  "name": "non-peer-deps",
+  "name": "fail-missing-deps",
   "description": "https://github.com/ng-packagr/ng-packagr/pull/687",
   "dependencies": {
     "foobar": "1.0.0"
   },
   "ngPackage": {
     "dest": "./dist",
+    "deleteDestPath": true,
```
alan-agius4, line 9: Can you explain why this is needed?

SchnWalter: When working on writing `fail*` integration sample tests, I've noticed that it's best to have a clean slate every time the test was executed. But now that I've removed the other more problematic tests, we can probably revert this one change.

P.S. I still have 2 new `fail` tests on my local machine, the fixes I've tried break the watch mode so I've postponed them for a later PR. One of those tests if for testing cleanup on builds failing because of missing imports inside secondary entry points. - the current code doesn't delete the destination.

PR #1846, opened 2021-01-24
test: use POSIX compatible code for building integration samples

SchnWalter:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Makes the shell code compatible with more systems by using POSIX compatible code. Main changes:

* Disabled glob expansion in favor of a simple dir command, this fixed a bug where the find command was actually treated as a glob and not as an actual command.
* Toggling the fail on error flag to avoid using `|| true`, this fixed a bug where builds that were expected to fail could stop failing because of regressions in the build system and the script would not see this as a problem.
* Took the liberty and added a link to ShellCheck, which is a nice utility that's available on most system.

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/samples.sh
+++ b/integration/samples.sh
@@ -1,37 +1,63 @@
-#!/bin/bash
-set -e
+#!/bin/sh
+# NOTE: When making changes, make sure to run the script through ShellCheck!
+# @see https://github.com/koalaman/shellcheck
 
-FIND_COMMAND=""
+# Exit script on error.
+set -o errexit
+# Disable pathname expansion.
+set -o noglob
+# Error on unset (non-special) parameter expansion.
+set -o nounset
 
-if [[ $OSTYPE == "win32" ]]; then
-    FIND_COMMAND="dir /b /o:n /ad"
+DIR_NAME_LIST=""
+if [ "${OSTYPE:-}" = "win32" ]; then
+    DIR_NAME_LIST=$(dir /b /o:n /ad "integration/samples/")
 else
-    FIND_COMMAND="find ./integration/samples/* -maxdepth 0 -type d"
+    DIR_NAME_LIST=$(dir --format=vertical --escape "integration/samples/")
 fi
+printf "Processing integration/samples:\n%s" "${DIR_NAME_LIST}"
+PACKAGE_FILE_CANDIDATES="ng-packagr-api.js  ng-package.js  ng-package.json  package.json"
 
-for D in $FIND_COMMAND
-do
-    P_JS=${D}/ng-package.js
-    P_JSON=${D}/ng-package.json
-    P_API=${D}/ng-packagr-api.js
-    P=${D}/package.json
+for DIR_NAME in $DIR_NAME_LIST; do
+    PACKAGE_PATH="integration/samples/${DIR_NAME}"
+    printf "\n\nProcessing integration sample: %s\n" "${PACKAGE_PATH}"
 
-    if [ -f $P_API ]; then
-        echo "Building sample ${P_API}..."
-        node ${P_API}
-    elif [ -f $P_JS ]; then
-        echo "Building sample ${P_JS}..."
-        node dist/cli/main.js -p ${P_JS}
-    elif [ -f $P_JSON ]; then
-        echo "Building sample ${P_JSON}..."
-        node dist/cli/main.js -p ${P_JSON}
-    elif [ -f $P ]; then
-        echo "Building sample ${P}..."
-        if [ $P='failures' ]; then
-            node dist/cli/main.js -p ${P} || true
-        else
-            node dist/cli/main.js -p ${P}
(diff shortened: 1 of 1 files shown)


integration/samples.sh
```diff
@@ -1,37 +1,60 @@
...
-    elif [ -f $P_JS ]; then
-        echo "Building sample ${P_JS}..."
-        node dist/cli/main.js -p ${P_JS}
-    elif [ -f $P_JSON ]; then
-        echo "Building sample ${P_JSON}..."
-        node dist/cli/main.js -p ${P_JSON}
-    elif [ -f $P ]; then
-        echo "Building sample ${P}..."
-        if [ $P='failures' ]; then
-            node dist/cli/main.js -p ${P} || true
```
SchnWalter, line 31: FYI: If for whatever reason the integration sample build no longer failed, this script didn't care. With the current changes we handle this situation using:
```
printf "ERROR: Build was expected to fail!\n"
exit 1
```

integration/samples.sh
```diff
@@ -1,37 +1,60 @@
...
-        node ${P_API}
-    elif [ -f $P_JS ]; then
-        echo "Building sample ${P_JS}..."
-        node dist/cli/main.js -p ${P_JS}
-    elif [ -f $P_JSON ]; then
-        echo "Building sample ${P_JSON}..."
-        node dist/cli/main.js -p ${P_JSON}
-    elif [ -f $P ]; then
-        echo "Building sample ${P}..."
-        if [ $P='failures' ]; then
```
SchnWalter, line 30: FYI: This was also failing (as I mentioned in https://github.com/ng-packagr/ng-packagr/pull/1845#discussion_r563229441). 

In the current PR we switched to a `case`, since that's the only POSIX compatible way to do it (besides grep, awk and what not).
```

    # All integration samples starting with "fail" are expected to fail.
    EXPECTED_TO_FAIL=false
    case "${DIR_NAME}" in
        "fail"*)
            printf "NOTE: This next integration sample build is expected to fail!\n"
            EXPECTED_TO_FAIL=true
            ;;
    esac
```

integration/samples.sh
```diff
@@ -1,37 +1,60 @@
...
-            node dist/cli/main.js -p ${P} || true
-        else
-            node dist/cli/main.js -p ${P}
+#!/bin/sh
+# NOTE: When making changes, make sure to run the script through ShellCheck!
+# @see https://github.com/koalaman/shellcheck
```
SchnWalter, line 3: I hope that it's OK to add this link here. This tool is very helpful in writing proper code. It reads the script bang and makes sure that only compatible syntax is used and also checks for a lot of features that might not work for the target shell.

integration/samples.sh
```diff
@@ -1,37 +1,60 @@
-#!/bin/bash
-set -e
-
-FIND_COMMAND=""
-
-if [[ $OSTYPE == "win32" ]]; then
-    FIND_COMMAND="dir /b /o:n /ad"
```
SchnWalter, line 7: @alan-agius4, can you please check what's failing? I don't have access to the logs. It's probably the removal of this line, but I'm wondering why this is needed. What are the pipelines for Windows are using? Is it `MinGW`?

SchnWalter (commented):
I've added a few of FYI comments and 2 questions, one regarding the failing pipeline.

alan-agius4 (approved):
Thanks for this. Although in future it's better not to do any changes in the bash files. Ideally we should move these to a node script for a number of reason.

1) More cross-platform compatible, 
2) Easier to understand for new collaborators.

integration/samples.sh
alan-agius4: Yes it is

SchnWalter:
> Ideally we should move these to a node script for a number of reason.

@alan-agius4, I'll look into this when I have the resources. And I'll also look into add the ability to run a single integration test, it would help a lot during development.

integration/samples.sh
```diff
@@ -1,37 +1,63 @@
...
+set -o nounset
 
-if [[ $OSTYPE == "win32" ]]; then
-    FIND_COMMAND="dir /b /o:n /ad"
+DIR_NAME_LIST=""
+if [ "${OSTYPE:-}" = "win32" ]; then
+    DIR_NAME_LIST=$(dir /b /o:n /ad "integration/samples/")
 else
-    FIND_COMMAND="find ./integration/samples/* -maxdepth 0 -type d"
+    DIR_NAME_LIST=$(dir --format=vertical --escape "integration/samples/")
```
alan-agius4, line 16: I am having some issues with this locally. On OSX, dir doesn't exist.

SchnWalter: Checkout the new PR where I've switched the integrations build script to JS: https://github.com/ng-packagr/ng-packagr/pull/1853

PR #1849, opened 2021-01-31
fix: ensure license entry point is properly built

### LINKED ISSUE #1451: Secondary entrypoint named "license" not working
## Type of Issue

```
[x ] Bug Report
[ ] Feature Request
```

## Description

I'm using ng-packagr to create an npm package with several secondary entrypoints.

One of the entrypoints is called "license".

All the other secondary entry-points are correctly generated. Is "license" a forbidden word?

#### How To Reproduce

I created a sample repository to reproduce the problem:
https://github.com/vmcbaptista/ng-packgr-issue

If you execute npm run build:components, you will see that the generated package.json under dist/components/license stays like:

```json
{
  "ngPackage": {}
}
```

#### Expected Behaviour
Comparing with the working secondary entry points I expect that the package.json generated look like:
```json
{
  "main": "../bundles/components-license.umd.js",
  "module": "../fesm5/components-license.js",
  "es2015": "../fesm2015/components-license.js",
  "esm5": "../esm5/license/components-license.js",
  "esm2015": "../esm2015/license/components-license.js",
  "fesm5": "../fesm5/components-license.js",
  "fesm2015": "../fesm2015/components-license.js",
  "typings": "components-license.d.ts",
  "metadata": "components-license.metadata.json",
  "sideEffects": false,
  "name": "components/license"
}
```

#### Version Information

```
ng-packagr:            5.7.1
@angular/compiler:     8.2.14
rollup:                1.25.2
tsickle:               0.37.1
typescript:            3.5.3
```

SchnWalter:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

1. On case-insensitive file systems, the copyFile command for the "LICENSE" file used to delete the output for a license entry point directory causing the build output to be incomplete. This PR adds a test and then fixes the build (for #1451).

2. Renamed `integration/samples/issue-852/spec/` to `specs/`, so that the tests are discoverable.

```
[ ] Yes
[x] No
```

--- 

Fixes #1451


src/lib/ng-package/package.transform.ts
```diff
@@ -237,11 +237,14 @@ const writeNpmPackage = (pkgUri: string): Transform =>
   pipe(
     switchMap(async graph => {
       const { data } = graph.get(pkgUri);
-      const files = [`${data.src}/LICENSE`, `${data.src}/README.md`];
```
SchnWalter, line 240: @alan-agius4, a few days ago, in https://github.com/ng-packagr/ng-packagr/pull/1840, you've removed `CHANGELOG.md` from the list of files that are automatically copied into the destination. Was the change intentional? Or should I add it back and add a test for it?

alan-agius4: Since we don’t generate a changelog we can leave it as is unless someone complains.

src/lib/ng-package/package.transform.ts
```diff
@@ -237,11 +237,14 @@ const writeNpmPackage = (pkgUri: string): Transform =>
...
+      const srcFiles = [`${data.src}/LICENSE`, `${data.src}/README.md`];
 
-      for (const src of files) {
-        if (await exists(src)) {
-          await copyFile(src, path.join(data.dest, path.basename(src)));
+      for (const srcFile of srcFiles) {
+        if (await exists(srcFile)) {
+          const srcFileStats = await stat(srcFile);
+          if (srcFileStats.isFile()) {
```
alan-agius4, line 245: We can remove 2 extra IO here, by using a try/catch block

```ts
for (const srcFile of srcFiles) {
  let isFile = false;
  try {
    isFile = (await stat(srcFile)).isFile();
  } catch {}

  if(isFile) {
    await copyFile(srcFile, path.join(data.dest, path.basename(srcFile)));
  }
}
```

PR #1856, opened 2021-02-04
fix: remove optional dependency on tsickle

alan-agius4:


kyliau:
rationale: Angular FW no longer ships code that's compatible with Closure compiler to NPM, so there's little value supporting `annotateForClosureCompiler` in `ng-packagr`. This means we no longer need `tsickle` here.

PR #1861, opened 2021-02-08
refactor: update to use Node 12 APIs

alan-agius4:
BREAKING CHANGE:
Node.js version 10 will become EOL on 2021-04-30.
Angular CLI 12 will require Node.js 12.13+ or 14.15+. Node.js 12.13 and 14.15 are the first LTS releases for their respective majors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
     "css",
     "scss"
   ],
+  "engines": {
+    "node": ">=12.13.0"
+  },
   "author": "David Herges <david@spektrakel.de>",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


SchnWalter:
You can also remove the Node.js v9 checks and just use the `getColorDepth()` method directly:

    colors.enabled = process.stdout.getColorDepth() > 1;

Instead of using the intermediate `supportsColor` constant:

https://github.com/ng-packagr/ng-packagr/blob/e105afa2129ebc3b336cd2c46c2d3bf5be269b78/src/lib/utils/color.ts#L6-L14

PR #1862, opened 2021-02-08
refactor: remove read-pkg-up dependency

SchnWalter:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

Removed direct dependency on `read-pkg-up@5`, it adds 40 dependencies for users of `ng-packagr`, and this dependency can be replaced by a few lines of code.

NOTE: The project still has an indirect dev-dependency through `standard-version`, which is used for making releases.

```
[ ] Yes
[x] No
```

--- 

Closes #1407

### DIFF
--- a/src/cli/main.ts
+++ b/src/cli/main.ts
@@ -2,8 +2,7 @@
 
 import * as program from 'commander';
 import * as path from 'path';
-import * as readPkgUp from 'read-pkg-up';
-import { execute, build, version } from '../public_api';
+import { build, execute, version } from '../public_api';
 
 const DEFAULT_PROJECT_PATH = path.resolve(process.cwd(), 'ng-package.json');
 
@@ -26,11 +25,8 @@ program
     value ? path.resolve(value) : undefined,
   );
 
-const dir = path.dirname(module.filename);
-const pkg = readPkgUp.sync({ cwd: dir }).pkg;
-
 program.on('option:version', () => {
-  version(pkg);
+  version();
   process.exit(0);
 });
 
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "ora": "^5.1.0",
     "postcss": "^8.2.4",
     "postcss-url": "^10.1.1",
-    "read-pkg-up": "^5.0.0",
     "rimraf": "^3.0.0",
     "rollup": "^2.37.0",
     "rollup-plugin-sourcemaps": "^0.6.3",
--- a/src/lib/commands/version.command.ts
+++ b/src/lib/commands/version.command.ts
@@ -8,17 +8,9 @@ import { Command } from './command';
  *
  * @stable
  */
-export const version: Command<any, void> = (pkg: any) => {
-  let TSICKLE_VERSION = '';
-  try {
-    TSICKLE_VERSION = require('tsickle/package.json').version;
-  } catch {}
-
-  const NG_PACKAGR_VERSION = pkg ? pkg.version : 'unknown';
-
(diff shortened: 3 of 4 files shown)


alan-agius4:
Thanks for this, though I think it's cleaner to just require the `package.json` to retrieve the version.

src/lib/utils/find-up.ts
```diff
@@ -0,0 +1,31 @@
+import { existsSync } from 'fs';
```
alan-agius4, line 1: This can be removed entirely version command is updated to require the JSON manifest
```ts
import { VERSION as ROLLUP_VERSION } from 'rollup';
import { VERSION as COMPILER_VERSION } from '@angular/compiler';
import { version as TS_VERSION } from 'typescript';
import { Command } from './command';

/**
 * Prints version information.
 *
 * @stable
 */
export const version: Command<any, void> = (pkg: any) => {
  let TSICKLE_VERSION = '';
  try {
    TSICKLE_VERSION = require('tsickle/package.json').version;
  } catch {}

  console.log(`ng-packagr:            ` + require('../package.json').version);
  console.log(`@angular/compiler:     ` + COMPILER_VERSION.full);
  console.log(`rollup:                ` + ROLLUP_VERSION);
  console.log(`tsickle:               ` + TSICKLE_VERSION);
  console.log(`typescript:            ` + TS_VERSION);
};

```

src/cli/main.ts
```diff
@@ -26,10 +26,9 @@ program
     value ? path.resolve(value) : undefined,
   );
 
-const dir = path.dirname(module.filename);
-const pkg = readPkgUp.sync({ cwd: dir }).pkg;
-
 program.on('option:version', () => {
```
alan-agius4, line 29:
```
program.on('option:version', () => {
  version();
  process.exit(0);
});
```

scripts/postbuild.js
```diff
@@ -8,7 +9,8 @@ copyFileSync('src/ng-package.schema.json', 'dist/ng-package.schema.json', consta
 mkdirSync('dist/lib/ts/conf', { recursive: true });
 copyFileSync('src/lib/ts/conf/tsconfig.ngc.json', 'dist/lib/ts/conf/tsconfig.ngc.json', constants.COPYFILE_FICLONE);
 
-const packageJson = JSON.parse(readFileSync('package.json'));
+const packageJsonPath = resolve(process.cwd(), 'package.json');
+const packageJson = require(packageJsonPath);
```
alan-agius4, line 13: In the case since we are mutating the manifest, it's best to use `JSON.parse(readFileSync('package.json'));`.

src/lib/utils/find-up.ts
SchnWalter: The problem is that the path is relative to the current module (`.../lib/commands`) and since the `package.json` is copied into the destination in `postinstall` and the relative location changes.

Instead of using a relative path directly, we should pass it as an argument and also have a short comment, like:
```ts
  // NOTE: This path is correct only after postinstall.
  const packageJsonPath = path.resolve(module.path, '../package.json');
  version(packageJsonPath);
```

alan-agius4: That is the reason why I used `require` and not a `import`.

Require will not be analysed by the compiler, and hence it will not cause the build to fail because the package manifest path doesn’t exist doesn’t build time.

SchnWalter: Done.

alan-agius4 (approved):
Thanks

Issue #1872, opened 2021-02-11
New StylesheetProcessor implementation is extremly slow


skrtheboss:
## Type of Issue
The new implementation of the `StylesheetProcessor` using workers is extremly slow
```
[x] Bug Report
[ ] Feature Request
```

## Description
I have added a `console.time('StylesheetProcessor')` at the beginning of the `StylesheetProcessor.process` call and a `console.timeEnd('StylesheetProcessor')` before the css return statement, these are the results:

ng-packagr@11.1.4
```
- Compiling TypeScript sources through NGC
StylesheetProcessor: 12.902ms
StylesheetProcessor: 8.599ms
StylesheetProcessor: 8.728ms
StylesheetProcessor: 19.27ms
StylesheetProcessor: 13.624ms
StylesheetProcessor: 9.034ms
StylesheetProcessor: 10.043ms
StylesheetProcessor: 24.805ms
StylesheetProcessor: 28.633ms
StylesheetProcessor: 7.321ms
StylesheetProcessor: 44.461ms
StylesheetProcessor: 21.771ms
StylesheetProcessor: 19.549ms
StylesheetProcessor: 6.383ms
StylesheetProcessor: 13.473ms
StylesheetProcessor: 6.555ms
StylesheetProcessor: 21.906ms
StylesheetProcessor: 8.931ms
StylesheetProcessor: 8.84ms
StylesheetProcessor: 18.113ms
```

ng-packagr@11.2.0
```
- Compiling TypeScript sources through NGC
StylesheetProcessor: 1.396s
StylesheetProcessor: 1.438s
StylesheetProcessor: 1.391s
StylesheetProcessor: 1.408s
StylesheetProcessor: 1.354s
StylesheetProcessor: 1.374s
StylesheetProcessor: 1.383s
StylesheetProcessor: 1.435s
StylesheetProcessor: 1.945s
StylesheetProcessor: 2.004s
StylesheetProcessor: 1.703s
StylesheetProcessor: 1.619s
StylesheetProcessor: 1.657s
StylesheetProcessor: 1.484s
StylesheetProcessor: 1.579s
StylesheetProcessor: 1.456s
StylesheetProcessor: 1.438s
StylesheetProcessor: 1.362s
StylesheetProcessor: 1.368s
StylesheetProcessor: 1.428s
```

I think the problem is, that every time a file must be processed a worker is started, instead of reusing an existing one.

#### Version Information

```
@angular/*: 11.2.0
[...]

PR #1875, opened 2021-02-12
perf: use sync-rcp instead of execFileSync to transform async function to sync functions #1872

skrtheboss:
By using sync-rpc ( which is also used by the [babel-plugin-postcss](https://github.com/unlight/babel-plugin-postcss/blob/master/src/getcss.ts#L14-L15) package ), we increase the performance
over execFileSync calls, which are very expensive.

## I'm submitting a...

```
[x] Performance improvement
[ ] Bug Fix
[ ] Feature
```

## Checklist

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "rxjs": "^6.5.0",
     "sass": "^1.32.5",
     "stylus": "^0.54.8",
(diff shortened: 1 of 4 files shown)


alan-agius4:
Thanks for this

PR #1878, opened 2021-02-15
refactor: use Node 12 receiveMessageOnPort with Atomics to synchroniz…

alan-agius4:


src/lib/styles/stylesheet-processor.ts
```diff
@@ -42,10 +41,25 @@ export class StylesheetProcessor {
...
 
-    return css;
+      Atomics.wait(signal, 0, 0);
+
+      const { css, warnings } = receiveMessageOnPort(ioChannel.port2).message;
+
+      warnings.forEach(msg => log.warn(msg));
+      return css;
+    } catch (error) {
+      throw error;
```
JoostK, line 59: Such catch, much wow?

src/lib/styles/stylesheet-processor.ts
```diff
@@ -42,10 +41,25 @@ export class StylesheetProcessor {
       browserslistData: this.browserslistData,
     };
 
-    const { css, warnings } = this.styleSheetProcessorWorker(workerOptions);
+    const ioChannel = new MessageChannel();
 
-    warnings.forEach(msg => log.warn(msg));
+    try {
+    const signal = new Int32Array(new SharedArrayBuffer(4));
```
JoostK, line 47: Indentation seems off here

JoostK (commented):
I don't have any knowledge in this area so can't really review this, but just two nits.

src/lib/styles/stylesheet-processor.ts
alan-agius4: 😁

src/lib/styles/stylesheet-processor.ts
alan-agius4: Umm... strange the commit hook should have formatted this.

src/lib/styles/stylesheet-processor.ts
alan-agius4: Forgot that we can use finally without catch

PR #1888, opened 2021-02-19
refactor: renaming whitelistedNonPeerDependencies in favor of allowedNonPeerDependencies

devsull:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
[X] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

## Description

https://github.com/ng-packagr/ng-packagr/issues/1884

Moved the functionality of `whitelistedNonPeerDependencies` into `allowedNonPeerDependencies`. 

Updated the JSON schema to reflect that `whitelistedNonPeerDependencies` is a deprecated field, added `allowedNonPeerDependencies` to the schema.  Also updated the documentation where the phrasing "white list" appears.

Added a warning message if `whitelistedNonPeerDependencies` is detected in the ng-package.json that it is being deprecated.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -181,9 +181,15 @@ async function writePackageJson(
 
   // Verify non-peerDependencies as they can easily lead to duplicate installs or version conflicts
   // in the node_modules folder of an application
-  const whitelist = pkg.whitelistedNonPeerDependencies.map(value => new RegExp(value));
+  const allowedList = pkg.allowedNonPeerDependencies.map(value => new RegExp(value));
   try {
-    checkNonPeerDependencies(packageJson, 'dependencies', whitelist, spinner);
+    if(pkg.whitelistedNonPeerDependencies) {
+      spinner.warn(
+        colors.yellow(`'whitelistedNonPeerDependencies' is being deprecated, use 'allowedNonPeerDependencies' instead.`)
+      );
+    }
+    
+    checkNonPeerDependencies(packageJson, 'dependencies', allowedList, spinner);
   } catch (e) {
     await rimraf(entryPoint.destinationPath);
     throw e;
@@ -241,7 +247,7 @@ async function writePackageJson(
 function checkNonPeerDependencies(
   packageJson: Record<string, unknown>,
   property: string,
-  whitelist: RegExp[],
+  allowed: RegExp[],
   spinner: ora.Ora,
 ) {
   if (!packageJson[property]) {
@@ -249,15 +255,15 @@ function checkNonPeerDependencies(
   }
 
   for (const dep of Object.keys(packageJson[property])) {
-    if (whitelist.find(regex => regex.test(dep))) {
-      log.debug(`Dependency ${dep} is whitelisted in '${property}'`);
+    if (allowed.find(regex => regex.test(dep))) {
+      log.debug(`Dependency ${dep} is allowed in '${property}'`);
     } else {
       spinner.warn(
         colors.yellow(
           `Distributing npm packages with '${property}' is not recommended. Please consider adding ${dep} to 'peerDependencies' or remove it from '${property}'.`,
         ),
       );
-      throw new Error(`Dependency ${dep} must be explicitly whitelisted.`);
+      throw new Error(`Dependency ${dep} must be explicitly allowed.`);
     }
   }
 }
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -183,12 +183,6 @@ async function writePackageJson(
   // in the node_modules folder of an application
   const allowedList = pkg.allowedNonPeerDependencies.map(value => new RegExp(value));
   try {
-    if(pkg.whitelistedNonPeerDependencies) {
-      spinner.warn(
-        colors.yellow(`'whitelistedNonPeerDependencies' is being deprecated, use 'allowedNonPeerDependencies' instead.`)
-      );
-    }
-    
     checkNonPeerDependencies(packageJson, 'dependencies', allowedList, spinner);
   } catch (e) {
     await rimraf(entryPoint.destinationPath);
@@ -255,7 +249,7 @@ function checkNonPeerDependencies(
   }
 
   for (const dep of Object.keys(packageJson[property])) {
-    if (allowed.find(regex => regex.test(dep))) {
+    if (allowed.some(regex => regex.test(dep))) {
       log.debug(`Dependency ${dep} is allowed in '${property}'`);
     } else {
       spinner.warn(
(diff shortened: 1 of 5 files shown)


src/ng-package.schema.json
```diff
@@ -23,14 +23,22 @@
...
+    "allowedNonPeerDependencies": {
       "description": "A list of dependencies that are allowed in the 'dependencies' and 'devDependencies' section of package.json. Values in the list are regular expressions matched against npm package names.",
       "type": "array",
       "items": {
         "type": "string"
       },
       "default": ["tslib"]
     },
+    "whitelistedNonPeerDependencies": {
+      "description": "A deprecated alias for 'allowedNonPeerDependencies'",
```
alan-agius4, line 35:
```suggestion
      "x-deprecated": "Use \"allowedNonPeerDependencies\" instead.",
      "description": "A list of dependencies that are allowed in the 'dependencies' and 'devDependencies' section of package.json. Values in the list are regular expressions matched against npm package names.",
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -181,9 +181,15 @@ async function writePackageJson(
 
   // Verify non-peerDependencies as they can easily lead to duplicate installs or version conflicts
   // in the node_modules folder of an application
-  const whitelist = pkg.whitelistedNonPeerDependencies.map(value => new RegExp(value));
+  const allowedList = pkg.allowedNonPeerDependencies.map(value => new RegExp(value));
   try {
-    checkNonPeerDependencies(packageJson, 'dependencies', whitelist, spinner);
+    if(pkg.whitelistedNonPeerDependencies) {
```
alan-agius4, line 186: This is redundant when using `x-deprecated` in the schema. This is because when using this keyword a warning is displayed when a deprecated option is used.

src/ng-package.schema.json
```diff
@@ -23,14 +23,22 @@
...
       "default": false
     },
-    "whitelistedNonPeerDependencies": {
+    "allowedNonPeerDependencies": {
       "description": "A list of dependencies that are allowed in the 'dependencies' and 'devDependencies' section of package.json. Values in the list are regular expressions matched against npm package names.",
       "type": "array",
       "items": {
         "type": "string"
       },
       "default": ["tslib"]
```
alan-agius4, line 32: Since we always add tslib programatically we can remove this from here

src/lib/ng-package/package.ts
```diff
@@ -67,10 +67,22 @@ export class NgPackage {
...
+   */
   public get whitelistedNonPeerDependencies(): string[] {
+    return this.primary.$get('whitelistedNonPeerDependencies') as string[];
+  }
+
+  public get allowedNonPeerDependencies(): string[] {
     // XX: default array values from JSON schema not recognized
     const defValue = ['tslib'];
-    const value = (this.primary.$get('whitelistedNonPeerDependencies') as string[]) || defValue;
+
```
alan-agius4, line 80: I don't think this logic is correct here, `whitelistedNonPeerDependencies` will always be truthy irrespective if it's being used or not.

I think something like the below should do
```
public get allowedNonPeerDependencies(): string[] {
  const allowedNonPeerDependencies = this.primary.$get('allowedNonPeerDependencies');
  if (this.primary.$get('allowedNonPeerDependencies').length) {
    return [...new Set(
      ...allowedNonPeerDependencies,
      'tslib',
    )];
  }

  return [...new Set(
    ...this.primary.$get('whitelistedNonPeerDependencies'),
    'tslib',
  )];
}
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -241,23 +247,23 @@ async function writePackageJson(
...
+    if (allowed.find(regex => regex.test(dep))) {
+      log.debug(`Dependency ${dep} is allowed in '${property}'`);
     } else {
       spinner.warn(
         colors.yellow(
           `Distributing npm packages with '${property}' is not recommended. Please consider adding ${dep} to 'peerDependencies' or remove it from '${property}'.`,
         ),
       );
-      throw new Error(`Dependency ${dep} must be explicitly whitelisted.`);
+      throw new Error(`Dependency ${dep} must be explicitly allowed.`);
```
alan-agius4, line 266:
```suggestion
      throw new Error(`Dependency ${dep} must be explicitly allowed using the "allowedNonPeerDependencies" option.`);
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -241,23 +247,23 @@ async function writePackageJson(
...
   spinner: ora.Ora,
 ) {
   if (!packageJson[property]) {
     return;
   }
 
   for (const dep of Object.keys(packageJson[property])) {
-    if (whitelist.find(regex => regex.test(dep))) {
-      log.debug(`Dependency ${dep} is whitelisted in '${property}'`);
+    if (allowed.find(regex => regex.test(dep))) {
```
alan-agius4, line 258:
```suggestion
    if (allowed.some(regex => regex.test(dep))) {
```

alan-agius4 (changes requested):
Thanks a bunch for this. 

Kindly see the comments below.

src/lib/ng-package/package.ts
devsull: Is it possible that the property coming from `$get()` could be `undefined`?  I like your changes on how the array is constructed, here's what I came up with:
```
  public get allowedNonPeerDependencies(): string[] {
    const alwaysInclude = ['tslib'];

    const allowedNonPeerDependencies = this.primary.$get('allowedNonPeerDependencies') as string[];
    
    // deprecated prop list
    const deprecated = this.whitelistedNonPeerDependencies;

    return Array.from(
      new Set([
        ...(deprecated ? deprecated : []),
        ...(allowedNonPeerDependencies ? allowedNonPeerDependencies :[]) ,
        ...alwaysInclude,
      ])
    );
  }
```
If it's true that this value will always be truthy then I can remove the trinary operators.

alan-agius4: The value cannot be undefined because a default is defined in the json schema.

devsull: Ah ha!  Ok, thank you for helping me understand.  Appreciate your code review - learning a lot :)

devsull:
This is my first public pull request, so please let me know how I can improve this process... I think there might be some processes I may not be following well!

src/lib/ng-package/package.ts
alan-agius4: I think merging both the whitelist and the allowed list here isn’t correct.

Maybe use the new option when provided instead of the deprecated one. Basically invert the logic in my example above.

alan-agius4 (approved):
One small comment, otherwise looks good!

Thanks again for your contribution. Welcome to the open source community 😀

src/lib/ng-package/package.ts
devsull: I've made that change

alan-agius4 (approved):

alan-agius4:
Thanks again

PR #1893, opened 2021-02-22
fix: removed default value from whitelistedNonPeerDependencies in json schema

devsull:
## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

Ref: https://github.com/ng-packagr/ng-packagr/issues/1892

Removed the default value from deprecated property `whitelistedNonPeerDependencies` in the json schema.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/lib/ng-package/package.ts
+++ b/src/lib/ng-package/package.ts
@@ -78,9 +78,9 @@ export class NgPackage {
     const alwaysInclude = ['tslib'];
 
     // remove in the future, after deprecation phase
-    const deprecatedPropList = this.whitelistedNonPeerDependencies ?? [];
+    const deprecatedPropList = this.whitelistedNonPeerDependencies;
 
-    if(deprecatedPropList.length) {
+    if(deprecatedPropList?.length) {
       return Array.from(
         new Set([
           ...deprecatedPropList,
--- a/src/lib/ng-package/package.ts
+++ b/src/lib/ng-package/package.ts
@@ -83,7 +83,7 @@ export class NgPackage {
     if(deprecatedPropList.length) {
       return Array.from(
         new Set([
-          ...deprecatedPropList,
+          ...(deprecatedPropList ? deprecatedPropList : []),
           ...alwaysInclude,
         ])
       );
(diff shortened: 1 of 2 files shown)


devsull:
It would be nice to write tests for this, but I'm not sure how to structure them, or test for these cases.... 😕

src/lib/ng-package/package.ts
```diff
@@ -78,12 +78,13 @@ export class NgPackage {
     const alwaysInclude = ['tslib'];
 
     // remove in the future, after deprecation phase
-    const deprecatedPropList = this.whitelistedNonPeerDependencies;
+    let deprecatedPropList = this.whitelistedNonPeerDependencies;
```
skrtheboss, line 81: Can be improoved: 

`const deprecatedPropList = this.whitelistedNonPeerDependencies ?? [];`

src/lib/ng-package/package.ts
```diff
@@ -78,7 +78,7 @@ export class NgPackage {
     const alwaysInclude = ['tslib'];
 
     // remove in the future, after deprecation phase
-    const deprecatedPropList = this.whitelistedNonPeerDependencies;
+    const deprecatedPropList = this.whitelistedNonPeerDependencies ?? [];
```
alan-agius4, line 81: Use the billable operator instead.

src/lib/ng-package/package.ts
```diff
@@ -78,7 +78,7 @@ export class NgPackage {
...
 
     if(deprecatedPropList.length) {
```
alan-agius4, line 83: `deprecatedPropList?.length`

alan-agius4:
Actually thinking about this, it’s a bug in the schema validation which is found https://github.com/ng-packagr/ng-packagr/blob/85eba9dab48d84fb591514807150cacb61068b16/src/lib/ng-package/schema.ts#L50

devsull:
@alan-agius4 I don't understand what the bug would be in what you mentioned above, about `schema.ts` ?

devsull:
> Actually thinking about this, it’s a bug in the schema validation which is found
> 
> https://github.com/ng-packagr/ng-packagr/blob/85eba9dab48d84fb591514807150cacb61068b16/src/lib/ng-package/schema.ts#L50
How is the bug here?

alan-agius4:
We should not display the warning when the value is defined by the user. 

That logic, is responsible for displaying the warning.

alan-agius4:
Took a look at this and there is no easy way to fix this in the schema registry.

PR #1905, opened 2021-03-06
fix: process only the typings files of packages with NGCC

alan-agius4:
Thanks to https://github.com/angular/angular/pull/40976 we can now process typings only


JoostK:
I think this would fail if `@angular/compiler-cli` < 11.2.4 is being used, as ngcc has configured `yargs` in strict mode so it errors for unknown flags.

alan-agius4:
And you are right

PR #1935, opened 2021-04-06
feat: allow publishing of Ivy library in partial mode.

### LINKED ISSUE #1901: Add configuration option to opt-out "ivy library" warning
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

Add an configuration option to disable the warning regarding publishing of "ivy libraries".

During build a message is shown warning to not publish libraries using ivy to NPM repositories. A project using only ivy components and encapsulation these into libraries is absolutely viable and should therefore not be warned about this during every build.

Warning get created and shown in `init-tsconfig.transform.ts`

#### How To Reproduce

Just use `ng build`

#### Expected Behaviour

Hide the warning, if the respective newly created configuration option opts out showing the warning.

#### Version Information

```
ng-packagr: 11.2.4
```

alan-agius4:
Ivy libraries are not allowed to be published when compiled in partial mode. To enable this mode, add `compilationMode: "partial"` under the `angularCompilerOptions` section in the library's tsconfig.

Closes #1901

### DIFF
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -11,6 +11,8 @@ import { globFiles } from '../../utils/glob';
 import { EntryPointNode, isEntryPointInProgress, isPackage, PackageNode, fileUrl } from '../nodes';
 import { Node } from '../../graph/node';
 
+type CompilationMode =  'partial' | 'full' | undefined;
+
 export const writePackageTransform: Transform = transformFromPromise(async graph => {
   const spinner = ora({
     hideCursor: false,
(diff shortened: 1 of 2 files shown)


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -10,6 +10,7 @@ import * as log from '../../utils/log';
 import { globFiles } from '../../utils/glob';
 import { EntryPointNode, isEntryPointInProgress, isPackage, PackageNode, fileUrl } from '../nodes';
 import { Node } from '../../graph/node';
+import { CompilationMode } from '@angular/compiler-cli/src/ngtsc/transform';
```
JoostK, line 13: I'd avoid this import and just stick to `'partial'|'full'` as that's how the option is defined. There is no guarantee that the internal enum has those exact values; in fact I don't expect this to work as the enum is numeric:

https://github.com/angular/angular/blob/3a823abcc554cffdda1e14e181dadad9063701de/packages/compiler-cli/src/ngtsc/transform/src/api.ts#L22-L32

JoostK (approved):

PR #1995, opened 2021-06-01
feat: update Ivy compilation pipeline to use faster NGTSC program

alan-agius4:


src/lib/ivy/compile-source-files.ts
```diff
@@ -0,0 +1,168 @@
+import * as ng from '@angular/compiler-cli';
```
alan-agius4, line 1: Most of the below, is coped over from https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/ivy/plugin.ts

integration/samples/api/tsconfig.ngc.json
```diff
@@ -1,11 +1,5 @@
 {
   "angularCompilerOptions": {
-    "enableIvy": false,
-    "annotateForClosureCompiler": false,
-    "skipTemplateCodegen": true,
-    "strictMetadataEmit": true,
-    "fullTemplateTypeCheck": true,
```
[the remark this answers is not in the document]
JoostK: Should we set `compilationMode: 'partial'` in all these tsconfigs?

src/lib/ivy/cache-compiler-host.ts
```diff
@@ -0,0 +1,166 @@
...
+
+      const cache = sourcesFileCache.getOrCreate(fileName);
+      if (cache.content === undefined) {
+        cache.content = compilerHost.readFile.call(this, fileName);
+      }
+      return cache.content;
+    },
+
+    // ng specific
+    moduleNameToFileName: (moduleName: string, containingFile: string) => {
```
JoostK, line 91: I don't think this should be implemented for ng-packagr:

https://github.com/angular/angular/blob/de93a7a4bb4fb79098c2d716f3380510ead2f62c/packages/compiler-cli/src/ngtsc/core/api/src/interfaces.ts#L11-L26

src/lib/ivy/cache-compiler-host.ts
```diff
@@ -0,0 +1,166 @@
...
+
+      return resourcePath;
+    },
+
+    readResource: async (fileName: string) => {
+      addDependee(fileName);
+
+      const cache = sourcesFileCache.getOrCreate(fileName);
+      if (cache.content === undefined) {
+        if (/(html|htm|svg)$/.test(path.extname(fileName))) {
```
JoostK, line 131: nit:

```suggestion
        if (/(?:html?|svg)$/.test(path.extname(fileName))) {
```

src/lib/ivy/cache-compiler-host.ts
```diff
@@ -0,0 +1,166 @@
...
+          // template
+          cache.content = compilerHost.readFile.call(this, fileName);
```
JoostK, line 133: Does this work? I was going to suggest using `this.readFile(fileName)` given that `this` appears to correspond with this object anyway, judging by the fact that it's passed to `.call(this, ...`. But since this is declared in an arrow function this has the lexical scope of `cacheCompilerHost` which would probably be the global object. I'm probably missing something as TypeScript is apparently not complaining.

src/lib/ivy/compile-source-files.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  let tsCompilerHost = cacheCompilerHost(
+    graph,
+    entryPoint,
+    tsConfigOptions,
+    moduleResolutionCache,
+    stylesheetProcessor,
+  );
+
+  if (tsConfigOptions.enableIvy && ngccProcessor) {
```
JoostK, line 35: This tests for `enableIvy`, but then below an `NgtscProgram` is always created which will therefore always be Ivy.

src/lib/ivy/compile-source-files.ts
```diff
@@ -0,0 +1,168 @@
...
+        builder.emit(sourceFile, undefined, undefined, undefined, transformers);
+      }
+    }
+  }
+
+  if (allDiagnostics.length === 0) {
+    return;
+  }
+
+  const exitCode = ng.exitCodeFromResult(allDiagnostics);
```
JoostK, line 161: This is duplicated above for `hasError`, so can probably be moved up.

src/lib/ivy/index.ts
```diff
@@ -0,0 +1,3 @@
+
```
JoostK, line 1: Unnecessary newline

src/lib/ng-package/entry-point/compile-ngc.transform.ts
```diff
@@ -24,31 +26,40 @@ export const compileNgcTransformFactory = (StylesheetProcessor: typeof Styleshee
...
       const { esm2015, declarations } = entryPoint.data.destinationFiles;
       const { basePath, cssUrl, styleIncludePaths } = entryPoint.data.entryPoint;
-      const { moduleResolutionCache, ngccProcessingCache, stylesheetProcessor = new StylesheetProcessor(basePath, cssUrl, styleIncludePaths) } = entryPoint.cache;
-      entryPoint.cache.stylesheetProcessor = stylesheetProcessor;
+      const { moduleResolutionCache, ngccProcessingCache } = entryPoint.cache;
 
-      const ngccProcessor = tsConfig.options.enableIvy
-        ? new NgccProcessor(ngccProcessingCache, tsConfig.project, tsConfig.options, entryPoints)
-        : undefined;
+      if (tsConfig.options.enableIvy !== false) {
```
JoostK, line 31: Maybe it would be wise to have an environment variable to disable the new pipeline; to allow people to fallback on the legacy pipeline if any issues pop up.

JoostK (commented):
I am a bit hesitant to start depending on internal compiler API, but since the CLI is already touching the same surface it shouldn't put ng-packagr in too much of a corner.

integration/samples/api/tsconfig.ngc.json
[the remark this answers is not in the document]
alan-agius4: > Should we set compilationMode: 'partial' in all these tsconfigs?

I think we should leave that to a seperate PR.

src/lib/ivy/cache-compiler-host.ts
alan-agius4: Yeah it does work.

src/lib/ng-package/entry-point/compile-ngc.transform.ts
alan-agius4: Yes agreeded.

src/lib/file-system/file-cache.ts
```diff
@@ -48,4 +47,14 @@ export class FileCache {
 
     return entry;
   }
+
+  updateAngularDiagnostics(sourceFile: ts.SourceFile, diagnostics: ts.Diagnostic[]): void {
+    if (this.has(sourceFile.fileName)) {
```
JoostK, line 52: I notice in the other methods that `this.normalizeKey` is used; we should probably do that here too such that deletions align with the keys used here.

src/lib/ivy/cache-compiler-host.ts
JoostK: This one was marked as resolved but the method is still here. Am I wrong in my suggestion that this should not be implemented?

src/lib/ivy/cache-compiler-host.ts
JoostK: This is also still left unchanged. Did some commit go missing maybe?

src/lib/ng-package/entry-point/compile-ngc.transform.ts
```diff
@@ -24,31 +33,40 @@ export const compileNgcTransformFactory = (StylesheetProcessor: typeof Styleshee
...
       const { esm2015, declarations } = entryPoint.data.destinationFiles;
       const { basePath, cssUrl, styleIncludePaths } = entryPoint.data.entryPoint;
-      const { moduleResolutionCache, ngccProcessingCache, stylesheetProcessor = new StylesheetProcessor(basePath, cssUrl, styleIncludePaths) } = entryPoint.cache;
-      entryPoint.cache.stylesheetProcessor = stylesheetProcessor;
+      const { moduleResolutionCache, ngccProcessingCache } = entryPoint.cache;
 
-      const ngccProcessor = tsConfig.options.enableIvy
-        ? new NgccProcessor(ngccProcessingCache, tsConfig.project, tsConfig.options, entryPoints)
-        : undefined;
+      if (tsConfig.options.enableIvy !== false && !isDisabled(process.env['NG_BUILD_LIB_LEGACY'])) {
```
JoostK, line 38: In my mind the environment variable is backwards. `NG_BUILD_LIB_LEGACY=1` implies you want the legacy pipeline, which lets `isDisabled` return `false`, becoming `true` after the negation and hence this code path is enabled when the `LEGACY` flag is set.

tl;dr; I think this should be
```suggestion
      if (tsConfig.options.enableIvy !== false && isDisabled(process.env['NG_BUILD_LIB_LEGACY'])) {
```

src/lib/ts/cache-compiler-host.ts
```diff
@@ -88,19 +89,6 @@ export function cacheCompilerHost(
       return cache.content;
     },
 
-    // ng specific
-    moduleNameToFileName: (moduleName: string, containingFile: string) => {
```
JoostK, line 92: Oh, it's gone from here. Ah, there's actually two copies of the compiler host and the legacy one was updated. I am surprised this method was here all along.

I now wonder how this could impact projects. With the `NG_BUILD_LIB_LEGACY` flag, would it be wise to keep the legacy host unchanged and only remove `moduleNameToFileName` from the new implementation, so that any consequences can be mitigated by enabling `NG_BUILD_LIB_LEGACY`.

src/lib/ivy/cache-compiler-host.ts
JoostK: I've figured it out, there's two compiler hosts! See my comment on the other one.

src/lib/file-system/file-cache.ts
alan-agius4: It’s redundant here,  since both `this.has` and `this.get` will call `normalizeKey` Internally

src/lib/ng-package/entry-point/compile-ngc.transform.ts
alan-agius4: You are toght

src/lib/file-system/file-cache.ts
JoostK: D'oh! Of course :-)

JoostK (approved):
The `html|htm|svg` regex is still in `src/lib/ivy/cache-compiler-host.ts`, otherwise LGTM

alan-agius4:
@JoostK updated.

PR #2003, opened 2021-06-09
fix: Debug Failure. False expression when using triple slash referenc…

alan-agius4:
…e directive

Array.fill last parameter is zero based `end` and not `length` of an array

## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/lib/ng-package/entry-point/analyse-sources.transform.ts
+++ b/src/lib/ng-package/entry-point/analyse-sources.transform.ts
@@ -58,11 +58,40 @@ function analyseEntryPoint(graph: BuildGraph, entryPoint: EntryPointNode, entryP
 
   const potentialDependencies = new Set<string>();
 
-  compilerHost.resolveModuleNames = (moduleNames: string[], containingFile: string) => {
-    if (containingFile.endsWith('.d.ts')) {
-      return [].fill(undefined, 0, moduleNames.length);
-    }
+  compilerHost.resolveTypeReferenceDirectives = (
+    moduleNames: string[],
+    containingFile: string,
+    redirectedReference: ts.ResolvedProjectReference | undefined,
+    options: ts.CompilerOptions,
+  ) => {
+    return moduleNames.map(moduleName => {
+      if (!moduleName.startsWith('.')) {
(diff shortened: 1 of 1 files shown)


src/lib/ng-package/entry-point/analyse-sources.transform.ts
```diff
@@ -60,7 +60,7 @@ function analyseEntryPoint(graph: BuildGraph, entryPoint: EntryPointNode, entryP
 
   compilerHost.resolveModuleNames = (moduleNames: string[], containingFile: string) => {
     if (containingFile.endsWith('.d.ts')) {
-      return [].fill(undefined, 0, moduleNames.length);
+      return [].fill(undefined, 0, moduleNames.length + 1);
```
JoostK, line 63: I am a bit surprised this is looking at `containingFile`, not `resolvedModule.fileName`. Wouldn't that mean that the first reference into a `.d.ts` file is considered but not subsequent imports, instead of all `.d.ts` files?

Additionally, maybe using `modulesNames.map(() => undefined)` would be more readable.

alan-agius4: Agreed with the map. 

I don't follow the argument around `resolvedModule.fileName`, at this point we don't want to resolve modules if the import is non relative.

src/lib/ng-package/entry-point/analyse-sources.transform.ts
```diff
@@ -80,7 +76,7 @@ function analyseEntryPoint(graph: BuildGraph, entryPoint: EntryPointNode, entryP
         moduleResolutionCache,
       );
 
-      return resolvedModule;
+      return resolvedModule.isExternalLibraryImport ? undefined : resolvedModule;
```
JoostK, line 79: FYI, `isExternalLibraryImport` should only be true for libs from `node_modules`. The author's subpack import will likely not have this flag set.

alan-agius4: I am not fixing 1982, I am fixing this comment https://github.com/ng-packagr/ng-packagr/issues/1982#issuecomment-857802373 :P

PR #2043, opened 2021-08-10
fix: only watch dependent files

### LINKED ISSUE #1723: Watch Dependencies and rebuild
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

The "ng build --watch" is not rebuilding when a dependency is rebuilt (another angular library).

We have a setup in which we have 2 libraries (Shared and Product) and 1 Angular project.

Shared library is used by both Angular and Product Library. It is shared from local dist.
```
package.json of library:
{
  "name": "product-lib",
  "version": "0.0.1",
  "peerDependencies": {
    "@angular/common": "^8.1.3",
    "@angular/core": "^8.1.3"
  },
  "dependencies": {
     "shared-lib":"file:../shared-proj/dist/shared-lib"
  }
}

package.json of Angular Project:
{
  "dependencies": {
     "shared-lib":"file:../shared-proj/dist/shared-lib"
  }
}
```

We try to do simultaneous development among the 3.
1. ng build --watch (shared-lib)
2. ng build --watch (product-lib)
3. ng serve (Angular)

On any change in the shared-lib the watch is initiated and the library is built.
Angular realizes that the library got rebuilt and it rebuilds itself with the library changes.
But product-lib is not able to detect the changes in its dependencies.

To make this work we have to stop the watch, "npm install" again and then start the watch.

#### How To Reproduce

Explained Above.

#### Expected Behaviour

The watch should also detect changes in dependencies and reinstall dep & rebuild itself similar to how Angular does it.

#### Version Information

```
ng-packagr: 10.0.4
@angular/compiler: 8.2.14
rollup:  2.24.0
tsickle:  0.35.0
typescript:  3.4.5
node: 12.7.0
npm: 6.10.0
```

If this is already supported please let me know how to.

Thanks

### LINKED ISSUE #1829: Allow user to add custom ignoredPaths for watcher
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description

I would like to pass a RegExp of files to ignore for the watcher. We have created a script for our monorepo that now controls the build order to avoid unnecessary rebuilds. But to have that control I need to ignore the dist folder for one of the two local dependencies for out package that uses ngpackgr. Now ng-packagr is triggered twice because one of it's dependencies are dependant on it's other dependency.

#### Expected Behaviour

The easiest solution would be to add it as a NgPackagrOptions and then pass it down to `createFileWatch` which already supports ignoredPaths` as an array of strings or RegExp.
image

The cleanest solution would probably be to add an option to the watch() call, but it would take more time to fix.

image

#### Version Information

```
ng-packagr: 11.0.3,
@angular/*: 11.0.x
typescript: 4.0.5
rxjs: 6.6.3
node: 14.x.x
npm/yarn: 6.x.x
```

_Please include all version numbers that might be relevant, e.g. third-party libraries_

alan-agius4:
With this change, we now only watch for files which are dependencies in the compilation. Previously, we watch the entire directory.

Closes #1829, closes 2042 and closes #1723


src/lib/file-system/file-watcher.ts
```diff
@@ -38,8 +41,11 @@ export function createFileWatch(
     });
   };
 
-  return Observable.create((observer) => {
-    watch.on('all', (event: AllFileWatchEvents, filePath: string) => handleFileChange(event, filePath, observer));
-    return () => watch.close();
-  });
+  return {
+    watcher: watch,
+    onFileChange: Observable.create((observer) => {
```
stephenlautier, line 46:
```suggestion
    onFileChange: new Observable((observer) => {
```
The other is deprecated 😄  ❤️

alan-agius4: Haha, at the moment the Linux failure is more serious than this deprecation.

PR #2050, opened 2021-08-20
feat: do not generate UMDs

### LINKED ISSUE #1674: Ability to reuse umdModuleId definitions in secondary entry points

### LINKED ISSUE #1757: Confusing error "WARNING: No name was provided for external module 'xxx' in output.globals – guessing 'xxx'" for absolute imports
## Type of Issue

```
[x] Bug Report
[x] Feature Request
```

## Description

When you use absolute imports inside lib you will get an absolutely confusing warning "No name was provided for external module 'xxx'". 

#### Desired Behaviour

1. Add supports for absolute imports (related to #916). IMO this is much much better because 
    - angular supported it
    - in a big app 'path/to/Component' much more readable than `../../../../../../../../../../../../Component`
    - mixed style is also confusing
2. Show smart warning eg "Absolute imports don't support, please use relative."

#### Version Information

```
ng-packagr:            10.1.2
@angular/compiler:     10.1.3
rollup:                2.28.2
tsickle:
typescript:            4.0.3
```

### LINKED ISSUE #2000: Generated code for built library is faulty
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

Building a library with Ivy either disabled or enabled results in wrong generated code.

```
function _interopNamespace(e) {
        if (e && e.__esModule) return e;
        var n = Object.create(null);
        if (e) {
            Object.keys(e).forEach(function (k) {
                if (k !== 'default') {
                    var d = Object.getOwnPropertyDescriptor(e, k);
                    Object.defineProperty(n, k, d.get ? d : {
                        enumerable: true,
                        get: function () {
                            return e[k];
                        }
                    });
                }
            });
        }
        n['default'] = e; // <----- set default property
        return Object.freeze(n);
}

var i0__namespace = /*#__PURE__*/_interopNamespace(i0);
    
// namespace called without default property
ModuleaService.ɵprov = i0__namespace.ɵɵdefineInjectable(...);
```

#### How To Reproduce

Minimal reproduction is available [here](https://github.com/redvp/ngplugins) projects/ngpluginsrepro is an app that loads the generated modules from the projects/plugins workspace. Generated modules are included into the app's assets.
You will find 2 working examples in the app's assets, modulea.umd.js and moduleb.umd.js. However, I cannot build the modules anymore. (Reverting to [this commit](ec1b97241453f265b934e6ae68306c93298d8318) and reinstalling node_modules doesn't help).

#### Expected Behaviour

I cannot explain the inconsistency, but I used to be able to build the libraries. The ngpluginsrepro app would then load these modules at runtime and compile them as Angular modules (also includes lazy-loading and routing to these modules).

#### Version Information

```
ng-packagr: 10.1.2
@angular/*: 10.1.6
typescript: 4.0.7
rxjs: 6.6.7
node: 12.18.3
npm/yarn: 6.14.6 / 1.22.10
```

### LINKED ISSUE #2023: When building in --watch mode, UMD bundle isn't created, yet still referenced
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description

while watching a library with `ng build --watch` the UMD bundle isn't created (as evident [here](https://github.com/ng-packagr/ng-packagr/blob/6576b9a3db69c158409550c1abbf03b484070a08/src/lib/ng-package/entry-point/write-bundles.transform.ts#L52), following this [commit](https://github.com/ng-packagr/ng-packagr/pull/1931)).

Yet, inside the generated `package.json` the `"main"` property still refers to the UMD bundle.

#### How To Reproduce

```
ng new test-umd
cd test-umd
ng generate library my-lib
ng build my-lib --watch
```

Now check `dist/my-lib/package.json`. You'll notice this line `"main": "bundles/my-lib.umd.js"`, while there is actually no `bundles` folder generated. (if you drop the `--watch` flag the folder IS generated).

#### Expected Behaviour

I expect that if the UMD bundling is skipped (as it is in `--watch`), refrain from pointing to a missing file in the `"main"` property of package.json.

My even bigger dream is that dropping the UMD bundling will be switchable. If a PR might help, I'll be happy if you could point me in the right direction.

#### Version Information

```
ng-packagr:            12.1.0
@angular/compiler:     12.1.1
rollup:                2.52.6
typescript:            4.3.5
```

alan-agius4:
BREAKING CHANGE:

We no longer generate UMD bundles. The below options which were used for UMD bundle generation has also been removed;

- `umdModuleIds`
- `amdId`
- `umdId`

Closes #2023, closes #2000, closes #1757 and closes #1674


SchnWalter:
Will this land on v13?

alan-agius4:
Yes

PR #2057, opened 2021-08-24
feat: replace es2015 with es2020

alan-agius4:
With this change we replace the ES2015 bundles with ES2020 variants.


SchnWalter:
Are there any other big changes like this one and https://github.com/ng-packagr/ng-packagr/pull/2050 planned for the next release?

SchnWalter:
I'm asking because on https://github.com/angular/angular/issues/38366, @alxhub put emphasis on ES2015 and that's a big jump since last year.

> As part of upgrading the APF, we're also proposing standardizing on ES2015 as a publication format.
> [...]
> Using ES2015 as the "source format of truth" strikes a balance between flexibility for translation into other needed formats, and ease of use with most tooling and browsers able to consume ES2015 without modification (so post-processing isn't required).

alan-agius4:
That comment is related to the ng-packagr distributable and not Angular libraries.

Basically the changes to drop UMDs and change the main output to ES2020 are part of the APF v13 spec which is currently still work in progress. One other change is that apart from FESM 2020, an additional FESM which contains ES2015 syntax (For legacy reasons) will be included.

The main reason for going with ES2020 is that most browsers now support ES2020 syntax without the need to downlevel to ES2015.

PR #2096, opened 2021-10-04
docs: remove section on UMD module names

JoostK:
The UMD bundles have been removed in APF13 so this commit the UMD module name configuration from the documentation.


alan-agius4 (approved):
Thanks @JoostK

PR #2152, opened 2021-11-11
docs: add details about exporting styles for APF 13

michaelfaith:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

This change adds a new documentation page detailing how to export any style assets that should be part of a library's public API.  This represents a change in how styles are treated since APF 13.

Discussed in https://github.com/ng-packagr/ng-packagr/issues/2149

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -97,6 +97,7 @@ Create one `package.json` per npm package, run _ng-packagr_ for each!
 - [Override tsconfig](docs/override-tsconfig.md)
 - [Add Style Include Paths](docs/style-include-paths.md)
 - [Package Secondary Entrypoints (sub packages)](docs/secondary-entrypoints.md)
+- [Export Styles (APF v13+)](docs/export-styles.md)
```
alan-agius4, line 100: The docs here are only for v13, hence no v13 is implied.

docs/export-styles.md
```diff
@@ -0,0 +1,54 @@
+# Export Styles
```
alan-agius4, line 1: I am thinking maybe we should just link to https://angular.io/guide/creating-libraries#managing-assets-in-a-library.

michaelfaith: Yeah, that makes sense.  Probably don't need the link on the README in that case.  Just linking from the Assets doc might be enough?  But it does become a bit circular then.  That page links to the assets doc in ng-packagr, and if we link back to it for the exports portion, that be strange.  What do you think?

michaelfaith:
I updated it, removing the references to v13, and consolidated the exports info into the Copy Assets doc as an addendum.  Still not sure if linking that to the `managing assets` doc on angular.io makes sense, since that page links to this document.  It could have people going in circles.  Thoughts?

docs/copy-assets.md
```diff
@@ -21,3 +21,55 @@ You can copy these assets by using the `assets` option.
...
+  "peerDependencies": {
+    ...
+  },
+  "dependencies": {
+    ...
+  }
+}
+```
+
+### Example ngPackage Config
```
alan-agius4, line 61: Can we remove the below? The type of config in package.json is deprecated.

docs/copy-assets.md
```diff
@@ -21,3 +21,55 @@ You can copy these assets by using the `assets` option.
...
+  "name": "your-library",
+  "version": "1.2.3",
+  "exports": {
+    ".": {
+      "sass": "./_index.scss"
+    },
+    "./styles/dark-theme": {
+      "sass": "./styles/_dark-theme.scss"
+    },
+    "./styles/_dark-theme": {
```
alan-agius4, line 42: Can we remove all the exports with `_`? These are due to maintain legacy backwards compatibility in Material but shouldn’t be needed otherwise.

docs/copy-assets.md
```diff
@@ -21,3 +21,55 @@ You can copy these assets by using the `assets` option.
   }
 }
 ```
+
+## Exporting Styles
+If your copied assets include styles that should be part of your library's public API, they need to be represented in the [`exports`](https://angular.io/guide/angular-package-format#exports) of its `package.json`.
+
+To do this, you can enumerate all such exports in an `"exports"` object in your base `package.json`.  `ng-packagr` will augment this and merge the additional exports generated from the build.
```
alan-agius4, line 28:
```suggestion
When including additional assets like Sass mixins or pre-compiled CSS. You need to add these manually to the conditional "exports" in the package.json of the primary entrypoint.

When including additional assets like Sass mixins or pre-compiled CSS. You need to add these manually to the conditional "exports" in the package.json of the primary entrypoint.

ng-packagr will merge handwritten "exports" with the auto-generated ones, allowing for library authors to configure additional export subpaths, or custom conditions.

```

docs/copy-assets.md
```diff
@@ -21,3 +21,55 @@ You can copy these assets by using the `assets` option.
   }
 }
 ```
+
+## Exporting Styles
+If your copied assets include styles that should be part of your library's public API, they need to be represented in the [`exports`](https://angular.io/guide/angular-package-format#exports) of its `package.json`.
```
alan-agius4, line 26:
```suggestion
```

michaelfaith:
Made the updates you suggested; that's a lot cleaner.  Appreciate the feedback.

docs/copy-assets.md
alan-agius4: Looks like you missed this comment.

michaelfaith: Ah, yep.  Sorry, didn't see that one for some reason.  I'll take care of that now.

michaelfaith: All set

alan-agius4 (approved):
Thanks for this.

PR #2164, opened 2021-11-18
Revert "build: update dependency postcss-preset-env to v7"

alan-agius4:
Reverts ng-packagr/ng-packagr#2159


alan-agius4:
Reverted due to licensing issues of  postcss-values-parser@6.0.1

PR #2280, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/graph/build-graph.ts
+++ b/src/lib/graph/build-graph.ts
@@ -46,7 +46,7 @@ export class BuildGraph implements Traversable<Node> {
     }
 
     if (this.watcher && node.url.startsWith(URL_PROTOCOL_FILE)) {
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):
Thanks

PR #2284, opened 2022-03-23
feat: enhance asset pattern

### LINKED ISSUE #1497: Asset copy Enhancement
## Type of Issue

```
[ ] Bug Report
[x] Feature Request
```

## Description
Recent PR https://github.com/ng-packagr/ng-packagr/pull/1465/ 
Introduced the feature to copy assets.
But the current format will limit us to :
1. Having all assets under project src .. ( from looking at the code)
  we are expecting assets to be under ::
  ```projects/<lib-name>/**/assets```
```<dist-path>/<lib-name>/**/assets```

with this i would not be able to use this ng-packagr feature as is ..

#### Expected Behaviour

I was looking at the AsstePattern syntax of anuglar-builder ( browser) .. https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/browser/schema.json#L393
```
export type AssetPattern =
  | {
      /**
       * The pattern to match.
       */
      glob: string;
      /**
       */
      input: string;
      /**
       */
      ignore?: string[];
      /**
       */
      output: string;
    }
  | string;
```

I wanted this feature badly for my projects so did a poc for the same ,in the forked repo https://github.com/arvindnama/ng-packagr/commit/957c8f057636d86bf7e1637fa567540bed7a81f3

NOTE: I will create a PR on the master branch with the above changes soon.

#### Version Information

```
ng-packagr: 9.0.0-rc3
@angular/*: 8.2.3
typescript: 3.5.3
rxjs: 6.5.3
node:12.13.1
npm 6.12.1
```

_Please include all version numbers that might be relevant, e.g. third-party libraries_

### PR DESCRIPTION
**Pull request by TheNightmareX**
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

Update of #1500 by @arvindnama

Related issue: #1497

> Recent PR https://github.com/ng-packagr/ng-packagr/pull/1465
> Introduced the feature to copy assets.
> But the current format will limit us to :
>
> 1.  Having all assets under project src .. ( from looking at the code)
> 2.  and we cannot customize the destination folder structure.
>     i.e. we are expecting assets to be under ::
>     projects/<lib-name>/**/assets
>     and destination will always be mapped to
>     <dist-path>/<lib-name>/**/assets
>
> in my project , we have multiple projects (libraries & apps)in same repo
> and we have a common folder for assets .. and this is out side of the project root (as expected)
>
> with this i would not be able to use this ng-packagr feature as is ..
> Expected Behaviour
>
> I was looking at the AsstePattern syntax of anuglar-builder ( browser) .. https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/browser/>schema.json#L393
>
> ```ts
> export type AssetPattern =
>   | {
>       /**
>        * The pattern to match.
>        */
>       glob: string;
>       /**
>        * The input directory path in which to apply 'glob'. Defaults to the project root.
>        */
>       input: string;
>       /**
>        * An array of globs to ignore.
>        */
>       ignore?: string[];
>       /**
>        * Absolute path within the output.
>        */
>       output: string;
>     }
>   | string;
> ```
>
> this will allow to be more granular and as a user of the ng-packagr i can get alot of flexibility ..

```
[ ] Yes
[x] No
```

Closes #1497

### CODE REVIEW

File: src/lib/ng-package/package.ts [hunk 1]
```diff
@@ -88,3 +88,5 @@ export class NgPackage {
     return [this.primary, ...this.secondaries].find(entryPoint => entryPoint.moduleId === moduleId);
   }
 }
+
+export type AssetEntry = { glob: string; input: string; ignore?: string[]; output: string };
```
alan-agius4, line 92: Can you change this to an interface please since in this case this describes an object which interfaces are intended for.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -35,55 +35,81 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
-      const assetFiles: string[] = [];
-
       // COPY ASSET FILES TO DESTINATION
       spinner.start('Copying assets');
 
       try {
-        for (const asset of [...ngPackage.assets, 'LICENSE', '**/README.md']) {
-          let assetFullPath = path.join(ngPackage.src, asset);
-
+        type AssetCopyTask = [srcPath: string, destPath: string];
```
alan-agius4, line 42: Can you move this outside of the method? Maybe below the imports?

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -35,55 +35,81 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+        const assetCopyTasks: AssetCopyTask[] = [];
+
+        // Parse string assets and add to tasks
+        const assetGlobs = ngPackage.assets
```
alan-agius4, line 46: The variable name here is misleading, since these can be globs and non globs.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -35,55 +35,81 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
-              continue;
-            }
-
-            if (stats.isDirectory()) {
-              assetFullPath = path.join(assetFullPath, '**/*');
-            }
+            const stats = await stat(assetFullGlob);
+            const isFile = stats.isFile() || assetGlob === 'LICENSE';
+            const isDir = stats.isDirectory();
+            assetFullGlobIs = isFile ? 'file' : isDir ? 'dir' : null;
```
alan-agius4, line 56: We should just use `isFile` and `isDir`.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -35,55 +35,81 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
-          let assetFullPath = path.join(ngPackage.src, asset);
-
+        type AssetCopyTask = [srcPath: string, destPath: string];
+        const assetCopyTasks: AssetCopyTask[] = [];
+
+        // Parse string assets and add to tasks
+        const assetGlobs = ngPackage.assets
+          .filter((asset): asset is string => typeof asset === 'string')
+          .concat('LICENSE', '**/README.md');
+        for (const assetGlob of assetGlobs) {
```
alan-agius4, line 49: IMHO, the logic here can be simplified and make it easier to follow, if we normalize all entries to globs. Also we need to make sure that the files are within the project root. 

You can take https://github.com/angular/angular-cli/blob/80d486f1592b19a571dfdd847c4811a68679d0cd/packages/angular_devkit/build_angular/src/utils/normalize-asset-patterns.ts#L29-L84 and https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/utils/copy-assets.ts as an inspiration. 

Also, it would be nice if the logic is moved to a seperate file.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -35,55 +35,81 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+            path.resolve(ngPackage.dest, path.relative(ngPackage.src, filePath)),
+          ]);
+          assetCopyTasks.push(...tasks);
+        }
+
+        // Parse object assets and add to tasks
+        const assetEntries = ngPackage.assets.filter((item): item is AssetEntry => typeof item !== 'string');
+        for (const assetEntry of assetEntries) {
+          const srcDirAbsPath = path.join(ngPackage.src, assetEntry.input);
+          const destDirAbsPath = path.join(ngPackage.dest, assetEntry.output);
```
alan-agius4, line 84: Since now the output can be provided, we need to valid that this doesn't escape the `dist`.

alan-agius4 (changes requested):
Thanks for your contribution.

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Hi. I'm not really sure about "non globs". If that means filenames, I guess filenames are also valid globs?

src/lib/ng-package/package.ts
TheNightmareX: Done

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: This type has been removed in the new implementation

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Done

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Implemented but still in the same file because I'm not sure about how to name the new file.

alan-agius4: Looks better at first glance but I will review later during the day. 

One thing I noticed (maybe I skipped) is that you missed to add the validation to verify that the input files are within the project root.

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: Yes that is correct

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Yes. The original author @arvindnama requested this feature to allow common assets between multiple projects, which cannot be achived if we limit the input. I accept both ones, so it's up to you.

alan-agius4: We should definitely limit the input to only copy files within the project root. This is also the same behaviour of Angular CLI.

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+      const glob = path.join(ngPackage.src, item);
+      const isDir = await stat(glob)
+        .then(stats => stats.isDirectory())
+        .catch(() => false);
+      asset = { glob: isDir ? path.join(glob, '**/*') : glob, input: '.', output: '/' };
+    }
+
+    asset.input = path.join(ngPackage.src, asset.input);
+    asset.output = path.join(ngPackage.dest, asset.output);
+    asset.ignore = [...(asset.ignore ?? []), ...pathsIgnored].map(glob => path.join(asset.input, glob));
```
alan-agius4, line 109: There is no need to join these ignored with the input paths. 

```suggestion
    asset.ignore = [...(asset.ignore ?? []), ...pathsIgnored];
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+    '.gitkeep',
+    '**/.DS_Store',
+    '**/Thumbs.db',
+    '**/node_modules/**',
+    `${ngPackage.dest}/**`,
+  ];
+
+  const assets: AssetEntry[] = [];
+
+  for (const item of ngPackage.assets.concat('LICENSE', '**/README.md')) {
```
alan-agius4, line 94: Nit

```suggestion
  for (const item of[...ngPackage.assets, 'LICENSE', '**/README.md']) {
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+    let asset: AssetEntry;
+    if (typeof item == 'object') {
+      asset = item;
+      asset.glob = path.join(ngPackage.src, asset.input, asset.glob);
```
alan-agius4, line 98: This seems to be redundant. 

```suggestion
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+    } else {
+      const glob = path.join(ngPackage.src, item);
+      const isDir = await stat(glob)
+        .then(stats => stats.isDirectory())
+        .catch(() => false);
+      asset = { glob: isDir ? path.join(glob, '**/*') : glob, input: '.', output: '/' };
+    }
+
+    asset.input = path.join(ngPackage.src, asset.input);
```
alan-agius4, line 107:
```suggestion
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+      throw new Error('Cannot write assets to a location outside of the output path.');
+    }
+
+    assets.push(asset);
+  }
+
+  for (const asset of assets) {
+    const fileSrcPaths = await globFiles(asset.glob, {
+      cwd: asset.input,
+      ignore: asset.ignore,
```
alan-agius4, line 122:
```suggestion
      ignore: [...(asset.ignore ?? []), ...pathsIgnored],
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+    if (!isOutputValid) {
+      throw new Error('Cannot write assets to a location outside of the output path.');
+    }
+
+    assets.push(asset);
+  }
+
+  for (const asset of assets) {
+    const fileSrcPaths = await globFiles(asset.glob, {
+      cwd: asset.input,
```
alan-agius4, line 121:
```suggestion
      cwd: path.join(ngPackage.src, asset.input),
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+
+  const assets: AssetEntry[] = [];
+
+  for (const item of ngPackage.assets.concat('LICENSE', '**/README.md')) {
+    let asset: AssetEntry;
+    if (typeof item == 'object') {
+      asset = item;
+      asset.glob = path.join(ngPackage.src, asset.input, asset.glob);
+    } else {
+      const glob = path.join(ngPackage.src, item);
```
alan-agius4, line 100: This is actually not a glob, hence better rename to something more meaningful 

```suggestion
      const assetPath = path.join(ngPackage.src, item);
```

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+        .then(stats => stats.isDirectory())
+        .catch(() => false);
+      asset = { glob: isDir ? path.join(glob, '**/*') : glob, input: '.', output: '/' };
+    }
+
+    asset.input = path.join(ngPackage.src, asset.input);
+    asset.output = path.join(ngPackage.dest, asset.output);
+    asset.ignore = [...(asset.ignore ?? []), ...pathsIgnored].map(glob => path.join(asset.input, glob));
+
+    const isOutputValid = !path.relative(ngPackage.dest, asset.output).startsWith('..');
```
alan-agius4, line 111: Nice!

alan-agius4, line 111: Instead of using using not operator `!` multiple times, consider changing the variable name instead. `isOutputPathInvalid`

src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -128,6 +74,70 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+    let asset: AssetEntry;
+    if (typeof item == 'object') {
+      asset = item;
+      asset.glob = path.join(ngPackage.src, asset.input, asset.glob);
+    } else {
+      const glob = path.join(ngPackage.src, item);
+      const isDir = await stat(glob)
+        .then(stats => stats.isDirectory())
+        .catch(() => false);
+      asset = { glob: isDir ? path.join(glob, '**/*') : glob, input: '.', output: '/' };
```
alan-agius4, line 104: Nit, might be easier to read and adding a comment might be helpful future contributors that might not be aware the filenames are valid globs. 

```suggestion
asset.output = '';
if (isDir) {
  asset.glob = '**/*';
  asset.input = assetPath;
} else {
  // Filenames are their own glob
  asset.glob = basename(assetPath);
  asset.input = dirname(assetPath);
}
```

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Okay. Btw, Angular CLI allows asset inputs from `node_modules`, maybe we can consider that?

src/ng-package.schema.json
```diff
@@ -77,5 +77,40 @@
     }
   },
   "required": [],
-  "additionalProperties": false
+  "additionalProperties": false,
+  "definitions": {
+    "assetPattern": {
+      "oneOf": [
+        {
+          "type": "object",
```
alan-agius4, line 85: Nit: can you also add

```
"followSymlinks": {
  "type": "boolean",
  "default": false,
  "description": "Allow glob patterns to follow symlink directories. This allows subdirectories of the symlink to be searched."
},
```

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: I remember a convention somewhere. It says it's better to name boolean variables affirmatively. What do you think?

src/lib/ng-package/package.ts
```diff
@@ -88,3 +88,10 @@ export class NgPackage {
     return [this.primary, ...this.secondaries].find(entryPoint => entryPoint.moduleId === moduleId);
   }
 }
+
+export interface AssetEntry {
+  glob: string;
+  input: string;
+  ignore?: string[];
+  output: string;
+}
```
alan-agius4, line 97: This is actually redundant since we can use the types generated from the JSON schema.

```ts
import { AssetPattern } from '../../ng-package.schema';
```

```suggestion

```

src/lib/ng-package/package.ts
```diff
@@ -65,7 +65,7 @@ export class NgPackage {
     return this.primary.$get('keepLifecycleScripts') === true;
   }
 
-  public get assets(): string[] {
+  public get assets(): (string | AssetEntry)[] {
```
alan-agius4, line 68:
```suggestion
  public get assets(): AssetPattern {
```

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Hi, I really have no idea what else it can be, could you give me some examples?

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: `Invalid` is still an affirmation. But I am okay either way.

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: In this case `item` can be a glob, directory or file path, hence naming it `glob` doesn't seem appropriate. Therefore I recommand `assetPath`.

TheNightmareX: Oh but when I saw a variable named as `xxPath`, I would think it as a pure path, and also file paths and directories are globs too, so maybe we can consider another name? or add a comment?

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Without this statement, the result paths of the glob will be relative rather than absolute.

TheNightmareX: Without this statement, the result paths of the glob will be relative rather than absolute.

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: The glob must be absolute to make the result paths absolute, so we cannot make the glob a filename

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: I'm making all the paths absolute so that we won't need to consider which are relative and which are not.

src/lib/ng-package/entry-point/write-package.transform.ts
TheNightmareX: Yes this way is better, sorry for my previous unthinking replies. I have few experiences in submitting PRs :[

src/ng-package.schema.json
alan-agius4: Looks like you missed the default value.

TheNightmareX:
@alan-agius4 Hi. ng-packagr allows glob patterns in both the string and object format, while Angular Builder doesn't support Glob in the string format. It may be a little complex to use a similar implementation like Angular Builder's. The current implementation converts make all paths/globs absolute, so we don't need to care about whether the string asset is a glob. I guess it might be more suitable?

alan-agius4:
I believe the suggest change https://github.com/ng-packagr/ng-packagr/pull/2284#discussion_r835246468 should cover that. I’d prefer if we keep the same behaviour as the browser builder.

TheNightmareX:
If we keep the same behavior as the browser builder, this feature might be a breaking change and the docs will need to be updated, because `assets: ["./styles/**/*.scss"]` will no longer work.

TheNightmareX:
> I believe the suggest change [#2284 (comment)](https://github.com/ng-packagr/ng-packagr/pull/2284#discussion_r835246468) should cover that. I’d prefer if we keep the same behaviour as the browser builder to avoid confusion.

```ts
if (isDir) {
  asset.glob = '**/*';
  asset.input = assetPath;
} else {
  // Filenames are their own glob
  asset.glob = basename(assetPath);
  asset.input = dirname(assetPath);
}
```

If we still support glob patterns in string format assets:

Here if `assetPath` is `"./styles/**/*.scss"`, then `isDir` will be false, and the result `asset` object will be `{ glob: "*.scss", input: "./styles/**/", output: "" }`, where `asset.input` is not a valid directory path.

While in the current implementation, the result `asset` will be like `{ glob: "/.../<workspace_name>/projects/<lib_name>/styles/**/*.scss", input: "/.../<workspace_name>/projects/<lib_name>/", "output": "/.../<workspace_name>/dist/<lib_name>/" }`, and we won't worry about invalid input paths.

src/ng-package.schema.json
TheNightmareX: It throws `strict mode: default is ignored for: data7.followSymlinks` when testing, so I removed the it. I'll try to solve it.

TheNightmareX: Solved

TheNightmareX:
@alan-agius4 Hello! What do you think? Whether we should remove glob support in string assets?

alan-agius4:
Let's do the breaking change so that we have the same behaviour of the browser builds. Will add a migration at a later stage.

alan-agius4:
Thinking more about it let's continue handling the string globs since this format is not deprecated as it should be pretty trivial to support. 

```ts
if (isDir) {
    glob = '**/*';
    input = assetPath;
} else if (isFile) {
    // Filenames are their own glob
    asset.glob = basename(assetPath);
    asset.input = dirname(assetPath);
} else {
    glob = assetPath
    input = ngPackage.src
}
```

TheNightmareX:
If we'll continue handling string globs, I guess the current implementation is generally ok? It normalizes all asset patterns into asset entries with all paths/globs absolute. I think this might be easier to understand.

src/lib/ng-package/entry-point/write-package.transform.ts
alan-agius4: Looks like you missed this.

alan-agius4:
In general, it would be better to avoid absolute paths in Globs as on Windows paths would need to be turned into POSIX paths https://github.com/isaacs/node-glob#windows.

Normalizing the paths to absolutely paths when you can provide the context to node-glob seems like a redundant step to me especially when this can be brittle on Windows.

TheNightmareX:
Hi @alan-agius4, I just noticed that there are some conflicts, so I pull-rebased the recent changes in this repo to my fork, and force-pushed my fork. Is this the right way to resolve conflicts?

alan-agius4:
> Hi @alan-agius4, I just noticed that there are some conflicts, so I pull-rebased the recent changes in this repo to my fork, and force-pushed my fork. Is this the right way to resolve conflicts?

Yes, please use rebase.

TheNightmareX:
Implementation updated @alan-agius4

alan-agius4 (approved):
LGTM, thanks

TheNightmareX:
Oh... This is counted as only one contribution... a bit disappointing... ＞︿＜

PR #2317, opened 2022-05-05
feat: update generated output to APF 14

alan-agius4:
With this change we update ng-packagr to generate libraries with APF 14 output. APF 14 changes are;
- Secondary entrypoints no longer contain a `package.json`.
- The entrypoint of the `.d.ts` has been changed to `index.d.ts` so that TypeScript can resolve these without the need of a nested `package.json`.

BREAKING CHANGE: `entryFile` can no longer be named `index.ts` as it will conflict with the generated `index.d.ts`. Please rename


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -40,36 +40,40 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
...
+          ngEntryPoint,
+          ngPackage,
+          {
+            module: relativeUnixFromDestPath(destinationFiles.fesm2015),
+            es2020: relativeUnixFromDestPath(destinationFiles.fesm2020),
+            esm2020: relativeUnixFromDestPath(destinationFiles.esm2020),
+            fesm2020: relativeUnixFromDestPath(destinationFiles.fesm2020),
+            fesm2015: relativeUnixFromDestPath(destinationFiles.fesm2015),
+            typings: relativeUnixFromDestPath(destinationFiles.declarations),
+            exports: ngEntryPoint.isSecondaryEntryPoint ? undefined : generatePackageExports(ngEntryPoint, graph),
```
devversion, line 62:
```suggestion
            exports: generatePackageExports(ngEntryPoint, graph),
```

src/ng-package.schema.json
```diff
@@ -54,6 +54,7 @@
         "entryFile": {
           "description": "Entry file to the public API (default: `src/public_api.ts`).",
           "type": "string",
+          "pattern": "^(?!index\\.ts).*$",
```
devversion, line 57: I have one question: If people cannot have `index.ts` in the sources, does that cause any problems with path mappings? e.g. assume I have a library in my CLI project and an application. The application will depend on my local library and somewhere I have a `tsonfig.json` with `paths` mappings. e.g.

```json
"paths": [
  "my-lib": ["libs/my-lib/public_api.ts"]
]

devversion (approved):

src/ng-package.schema.json
alan-agius4: This is actually not recommended, in the Angular CLI that path mappings will always point to the built library.

alan-agius4: Screenshot 2022-05-05 at 11 59 28

https://angular.io/guide/creating-libraries

devversion: I see, that makes sense. sounds like we are good then!

PR #2426, opened 2022-09-08
refactor: remove `postcss-preset-env`

alan-agius4:
This change removes the usage of `postcss-preset-env` as this is no longer needed since Angular no longer supports browsers that require polyfills for CSS stage 3 features https://preset-env.cssdb.org/features/#stage-3.

We replace this and use autoprefixer directly which is still needed.

```
npx autoprefixer --info
Browsers:
  Chrome: 105
  Edge: 105, 104
  Firefox: 104, 102, 91
  iOS Safari: 15.6, 15.5, 15.4, 15.2-15.3, 15.0-15.1, 14.5-14.8, 14.0-14.4
  Safari: 15.6, 15.5, 15.4, 15.2-15.3, 15.1, 15, 14.1, 14

These browsers account for 20.38% of all users globally

At-Rules:
  @resolution: webkit

Selectors:
  ::backdrop: webkit
  ::file-selector-button: webkit
  :autofill: webkit
  :fullscreen: webkit

Properties:
  appearance: webkit
  backdrop-filter: webkit
  backface-visibility: webkit
  background-clip: webkit
  box-decoration-break: webkit
  color-adjust: webkit, moz
  hyphens: webkit
  mask-border-outset: webkit
  mask-border-repeat: webkit
  mask-border-slice: webkit
  mask-border-source: webkit
  mask-border-width: webkit
  mask-border: webkit
  mask-clip: webkit
  mask-composite: webkit
  mask-image: webkit
  mask-origin: webkit
  mask-position: webkit
  mask-repeat: webkit
  mask-size: webkit
  mask: webkit
  print-color-adjust: webkit, moz
  text-decoration-color: webkit
  text-decoration-line: webkit
  text-decoration-skip-ink: webkit
  text-decoration-skip: webkit
  text-decoration-style: webkit
  text-decoration: webkit
  text-size-adjust: webkit
  user-select: webkit

Values:
  cross-fade: webkit
  element: moz
  fill-available: webkit
  fill: webkit
  fit-content: moz
  image-set: webkit
  isolate: webkit
  stretch: webkit, moz
```


romainmenke:
Hi there!

Obviously we don't mind that you drop `postcss-preset-env`.
The end goal of any polyfill or fallback is to become obsolete.

But wanted to add that "Stage 3" is a completely fictional construct.
We have tried to properly define the stages in cssdb, but ultimately it just doesn't exactly align with how features mature during the specification process and how they are rolled out by browser vendors.

There are many features in `postcss-preset-env` that overlap with the browsers that Angular does support (and more being added all the time). These might not have the Stage 3 flag at this time purely because we rarely update this.

A better indicator is the number of implementations : https://github.com/csstools/postcss-plugins/tree/main/plugin-packs/postcss-preset-env#minimumvendorimplementations

alan-agius4:
Hi @romainmenke, Thanks for the information. I was not aware that Stage was not kept in sync.

PR #2447, opened 2022-10-07
feat: drop support for TypeScript 4.6 and 4.7

alan-agius4:
This commit drops support for TypeScript 4.6 and 4.7

BREAKING CHANGE: TypeScript versions older than 4.8.2 are no longer supported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^15.0.0-next",
     "tslib": "^2.3.0",
-    "typescript": ">=4.6.2 <4.9"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^15.0.0-next",
     "tslib": "^2.3.0",
-    "typescript": ">=4.6.2 <4.9"
+    "typescript": "^4.8.2"
```
JoostK, line 65: Shouldn't this be `~4.8.2`?

alan-agius4: Oops

PR #2512, opened 2022-12-20
feat: add support for dynamic import

alan-agius4:
This commits add supports to process dynamic imports and output lazy chunks.

**NB:** This does not mean that lazy routing is a good idea to have in a library.  There are cases outside of routing why you might want to use dynamic imports.

Closes https://github.com/angular/angular-cli/issues/21358 and closes https://github.com/ng-packagr/ng-packagr/issues/2508


Carlosamouco:
Thank you so much for this.

PR #2789, opened 2024-02-14
fix: support string as plugin option in custom postcss plugin config

alan-agius4:
In certain cases, the plugin option may be a string value, as shown in the example below:

```json
{
 "plugins": {
  "tailwindcss/nesting": "postcss-nesting"
  }
}
```

See: https://tailwindcss.com/docs/using-with-preprocessors#nesting


src/lib/styles/postcss-configuration.ts
```diff
@@ -90,7 +88,7 @@ export async function loadPostcssConfiguration(
 
   const config: PostcssConfiguration = { plugins: [] };
   for (const [name, options] of entries) {
-    if (!options || typeof options !== 'object') {
+    if (!options || typeof options !== 'object' || typeof raw.plugins !== 'string') {
```
JoostK, line 91: I am a bit confused by this condition; should it be something like

```suggestion
    if (!options || (typeof options !== 'object' && typeof options  !== 'string')) {
```

alan-agius4: You are right

PR #2897, opened 2024-08-08
feat: add support for node module resolution in style @use and @import

skrtheboss:
Currently the import resolver does not work in pnpm repositories, since for example if you install @angular/material and import something from it, it will not find the @material/typography package, since it is not symlinked to the root node_modules directory. Instead it can be found relative to the @angular/material location, for example node_modules/.pnpm/@angular+material@version/node_modules/@material/typography. This commit add a FileImporter which follows the node module resolution, an therefore find the imports correctly.

## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

This PR introduces a new FileImporter class to handle node module resolution more effectively, particularly for projects using pnpm. Previously, the import resolver failed to locate packages installed via pnpm due to the absence of symlinks in the root node_modules directory. For example, importing a module from @angular/material would not resolve @material/typography correctly, as pnpm installs these modules in nested directories like node_modules/.pnpm/@angular+material@version/node_modules/@material/typography.

The new FileImporter class follows node module resolution rules and is capable of resolving these imports correctly by navigating the package structure appropriately.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/styles/stylesheet-processor-worker.ts
+++ b/src/lib/styles/stylesheet-processor-worker.ts
@@ -1,8 +1,10 @@
 import { build, formatMessages } from 'esbuild';
-import { dirname, extname, relative } from 'node:path';
-import { pathToFileURL } from 'node:url';
+import { existsSync, realpathSync } from 'node:fs';
+import { dirname, extname, join, parse, relative, resolve } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { workerData } from 'node:worker_threads';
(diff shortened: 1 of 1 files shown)


alan-agius4:
Hi @skrtheboss, thank you for your contribution. However, this has already been addressed as part of a larger overhaul.

#2941

skrtheboss:
Hi @alan-agius4 ,

Thanks for the update! Glad to hear it's already been addressed.

Issue #2375, opened 2022-07-08
Do not know how to serialize a BigInt


TomMettam:
## Type of Issue

```
[X] Bug Report
[ ] Feature Request
```

## Description

When building an angular library, I get the error "Do not know how to serialize a BigInt"

My project does use BigInts, but doesn't attempt to serialize them at any point

```
------------------------------------------------------------------------------
Building entry point '@genius/frontend'
------------------------------------------------------------------------------
✔ Compiling with Angular sources in Ivy full compilation mode.
✖ Generating FESM2020
Do not know how to serialize a BigInt
```

#### How To Reproduce

It's unclear exactly what is being serialized here so I don't know how to reproduce the error

#### Expected Behaviour

Project should build

#### Version Information

```
$ node_modules/.bin/ng-packagr --version
ng-packagr:            14.0.3
@angular/compiler:     14.0.5
rollup:                2.75.7
typescript:            4.7.4
```

_Please include all version numbers that might be relevant, e.g. third-party libraries_

alan-agius4:
Hi @TomMettam,

I am unable to replicate. Can you please provide a minimal reproduction?

alan-agius4:
I suspected that it might be rollup, but without a reproduction it is hard to tell what is happened.

TomMettam:
Hello.

Creating a minimal reproduction isn't feasible for me because my project is huge, and the error reported from ng-packagr isn't in any way helpful in knowing what might be triggering the errror. So I added some code to diagnose the issue.  

The exception is thrown in `rollup.js` with the following line:

`await (0, cache_1.saveCacheEntry)(cacheDirectory, opts.cacheKey, JSON.stringify(bundle.cache));`

I added some code to scan the object for bigints, and it spat out the following:


```
modules[3].ast.body[7].declaration.body.body[0].value.body.body[0].alternate.alternate.alternate.alternate.alternate.alternate.alternate.consequent.body[0].alternate.consequent.body[0].alternate.alternate.alternate.body[1].test.right.argument.value
bigint found at modules[3].ast.body[7].declaration.body.body[0].value.body.body[0].alternate.alternate.alternate.alternate.alternate.alternate.alternate.consequent.body[0].alternate.alternate.body[0].alternate.alternate.alternate.body[1].test.right.value
[...]

PR #2931, opened 2024-10-15
fix: improve cache logic to handle BigInt

alan-agius4:
Closes: #2375


src/lib/utils/cache.ts
```diff
@@ -18,22 +18,56 @@ export async function generateKey(...valuesToConsider: string[]): Promise<string
...
 
-export async function saveCacheEntry(cachePath: string, key: string, content: string): Promise<void> {
-  await cacache.put(cachePath, key, content);
+export async function saveCacheEntry(cachePath: string, key: string, content: any): Promise<void> {
+  const filePath = join(cachePath, key);
+
+  // Ensure the cache directory exists
+  await ensureCacheDirExists(cachePath);
+
+  const data = JSON.stringify(content, (_key, value) => (typeof value === 'bigint' ? value.toString() + 'n' : value));
```
JoostK, line 70: WDYT of making this a bit more explicit, like `%BigInt(${value})` to reduce the chance of stripping off an `n` suffix of a string that happens to be `(\d+)n`

alan-agius4: Good idea.

src/lib/utils/cache.ts
```diff
@@ -11,29 +11,65 @@ try {
   ngPackagrVersion = require('../../../package.json').version;
 }
 
+const BIGINT_STRING_VALUE_REGEXP = /%BigInt\((\d+)\)/;
```
JoostK, line 14:
```suggestion
const BIGINT_STRING_VALUE_REGEXP = /^%BigInt\((\d+)\)$/;
```

JoostK (commented):
I think this should work, just one suggestion to anchor the regex to reduce its search space.

JoostK (approved):

Issue #2936, opened 2024-10-21
Optimize entry points rebuilds


Carlosamouco:
## Type of Issue

```
[ ] Bug Report
[X] Feature Request
```

## Description

When we are developing a library, it's important to pinpoint what are the changes and what needs to be rebuilt to save computational resources. This is especially important as a library grows, and the amount of code and the number of entry points increases. Also, rebuilding unnecessary entry points can trick other watchers making them rebuild, like an application builder, further degrading the development experience.

Currently, there are a few problems with having multiple entry points. For example, having a component for entry point results in a much slower build time than having all in one entry point. In a test that I did with 3 components, it takes 4 seconds to build 3 entry points vs 2 seconds to build a single-entry point with the same 3 components. I am not sure what could be done here, but maybe building nondependent entry points in parallel could help.

I also think we could improve the way we are rebuilding entry points in watch mode. When we change a source file, we are always assuming that it will produce side effects in its declaration file, and we end up rebuilding any entry points that depend on that declaration file. It does not make sense to always rebuild dependent entry points because we could be updating something that does not cause changes in the *d.ts files like changing a component’s template. It would also be nice if we could only build/watch a specific entry point in cases where we know that we will not be causing breaking changes, avoiding unnecessary recompilation of other entry points and avoiding triggering any application watcher.

All the described problems discourage people from building more optimized libraries where each component is exported in their own entry point, just like in @angular/material.

#### Version Information

```
$ node_modules/.bin/ng-packagr --version
ng-packagr:            19.0.0-next.6
@angular/compiler:     19.0.0-next.9
typescript:            5.6.3
```

PR #2937, opened 2024-10-21
fix: watch dist folder to detect changes in declaration files

Carlosamouco:
* rely on changes to the files in the dist folder when invalidating declaration files that an entrypoint depends on
* only invalidate the *.ts file of a component when updating its template or style file
* fixes problem where the template errors would not be shown when rebuilding a entry point
* entry point build pipelines are now correctly canceled when new changes are detected in watch mode

## I'm submitting a...

- [X] Bug Fix
- [ ] Feature

## Checklist

## Description

Currently, when in watch mode, if we change a *.ts file, the associated declaration file is always invalidated causing dependent entry points to be rebuilt.

Also if we change a template or a style file of a component, we should only invalidate the associated *.ts file of that component and not other files or entry points.

In order to optimize entry point rebuilds, libraries now watch also their own dist folder to make entry points rebuilds based on changes made to the emitted declaration files, instead of always invalidating the declaration files of the source file that was changed.

#2936

- [ ] Yes
- [X] No


alan-agius4 (commented):
Thank you for your contribution. However, I have a few concerns regarding this approach:

1. With the new logic, if one of the entry-point builds is slow (for example, due to a large number of assets being copied), the DTS file watching might cancel the build. This happens because the DTS emission occurs before the assets are processed.

2. While adding listeners to the `dist` folder could help, I’m concerned about introducing too many watchers, especially for large entry points, which could trigger a lot of unnecessary file change events.

I’ve explored alternatives in the past that achieve the same goal without relying on watching the `dist` folder. I ran into was that the TypeScript files weren’t invalidating correctly, which I did not solve it yet.

That being said, I’m not sure when I’ll be able to revisit it. However, I’m happy to share the code if you’d like to contribute.

Carlosamouco:
Thank you for the feedback,

Since the transformers fetch the entry point being built by searching for an entry point with state `in-progress`, if more than one entry point is in this `in-progress` state, weird issues can occur. For example, the ngcc transformer could be rebuilding the wrong entry point by mistake.

It’s important to be cautious when starting a pipeline with a different entry point when there is already another one being built. Currently there are already issues regarding this, sometimes you can see in the console errors like `ERROR: Cannot read properties of undefined (reading 'data')` when all of a sodden no entry points are in the ‘in-progress’ state, but its more problematic if a new entry point starts to be processed when there is another entry point in the 'in-progress' state.

I am not ensuring that a pipeline ends before triggering a new one and that can cause problems as you said. You also mentioned that the pipeline is canceled but I did not figure out how we are doing that.

About the number of listeners increasing, I thought that maybe it would be fine, since the application builder watches for many more files, including all the local libraries dest folder and in some cases the `node_modules` as well.

I am interested in hearing what the preferable approach would be so I can see if I can implement it. Maybe we need to send the entry point along side the graph to each transformer, so they don't get mixed when they change state and have a more efficient way to cancel a pipeline if new changes are pushed to an entry point.

Carlosamouco:
I was doing some tests and noticed that if an entry point fails to compile due to an error in a template, after triggering a couple of rebuilds just by saving that file without making any changes to it, the angular compiler stops reporting errors and the compilation of the entry point succeeds. If I delete the `cache.oldNgtscProgram` or the `cache.oldBuilder` it starts reporting errors again. Do you have any clue why this is happening? It only misbehaves with *.html files.

Carlosamouco:
I took a look at the angular compiler cli, and when they are in watch mode they don't reuse the old program more than once. I think we need to implement the same logic here.

PR #2942, opened 2024-10-27
fix: improve source files invalidation

Carlosamouco:
* only invalidate the *.ts file of a component when updating its template or style file
* fixes problem where the template errors would not be shown when rebuilding

## I'm submitting a...

- [X] Bug Fix
- [ ] Feature

## Checklist

## Description

If we change a template or a style file of a component, we should only invalidate the associated *.ts file of that component and not other files or entry points.

This also intends to fix a bug were when rebuilding a component template, the compiler would stop reporting diagnostics errors.

#2936

- [ ] Yes
- [X] No


src/lib/ts/cache-compiler-host.ts
```diff
@@ -289,3 +288,15 @@ export function augmentProgramWithVersioning(program: ts.Program): void {
     return files;
   };
 }
+
+export function invalidateProgramFiles(program: ts.Program, files: Set<string>): void {
```
alan-agius4, line 292: It's odd, since TypeScript should manage this automatically. I'm fairly sure we don’t handle this in the Angular CLI.

Could you provide a detailed explanation or reproduction steps to show why this is needed?

//cc @JoostK, as he has a deep understanding of TypeScript internals.

JoostK: I'd have to check some things to determine what is the case here. Is there a test that would show the issue and how these changes address it?

src/lib/ts/cache-compiler-host.ts
```diff
@@ -289,3 +288,15 @@ export function augmentProgramWithVersioning(program: ts.Program): void {
...
+  files.forEach(file => {
+    const source = program.getSourceFile(file) as ts.SourceFile & { version?: string };
+
+    if (source) {
+      source.version = createHash('sha256')
+        .update(source.text + new Date().getTime())
```
JoostK, line 298: I don't quite understand the intent here. Adding the current time as part of the digest would mean the source file's content is mostly irrelevant. In any case, it would be better to split the string concatenation into two independent `update` calls.

I'd have to check how the `ts.SourceFile.version` is treated internally for better insight into what this means exactly.

src/lib/ts/cache-compiler-host.ts
Carlosamouco: I noticed that when you save a component file for a couple of times that as errors in a template, suddenly, the diagnostics stop being reported and the entry points gets built normally.

I think the problem is that we are relying on the `affectedFiles` to run the angular diagnostics, and when you re-save a file without changing its content, that file is not marked as affected anymore by the `getSemanticDiagnosticsOfNextAffectedFile`. 

`compile-source-files.ts`
```ts
// Only request Angular template diagnostics for affected files to avoid
// overhead of template diagnostics for unchanged files.
if (affectedFiles.has(sourceFile)) {
  const angularDiagnostics = angularCompiler.getDiagnosticsForFile(...);
  ...
}
```

I was making sure those files are re-versioned so they are picked up again by the builder but maybe I could change that condition in some way?

On a final note, in the angular compiler cli, the builder is not reused again when there are errors, I think.

alan-agius4: > I noticed that when you save a component file for a couple of times that as errors in a template, suddenly, the diagnostics stop being reported and the entry points gets built normally.

Oh that is expected and working as intended from TS point of view.  we should be storing and reusing previous errors and failures instead of invalidating the file unless it has actually changed.

> On a final note, in the angular compiler cli, the builder is not reused again when there are errors, I think.

The compiler CLI does not utilize TypeScript builders, and it just re-uses the old program.

Carlosamouco: Ok that makes sense, I will update my PR then. I am thinking in comparing the version of the previous affected files with the version of the matching current files, and if they are the same, I will just retrieve the old diagnostics.

Carlosamouco: I have updated the code. I am not wasting your time. :(

src/lib/ngc/compile-source-files.ts
```diff
@@ -185,7 +194,10 @@ export async function compileSourceFiles(
   }
 
   if (errorDiagnostics.length) {
+    entryPoint.cache.affectedFiles = new Set(affectedFiles);
```
alan-agius4, line 197: In this case, we should only cache the Angular diagnostics results, we should avoid re-running the diagnostics for files that did not actually change. 

See: https://github.com/angular/angular-cli/blob/11289c4982dfaada860a595fa386d6176f696322/packages/angular/build/src/tools/angular/compilation/aot-compilation.ts#L184

Carlosamouco: I am not really sure how to approach this. How do I know when the content of a template file has changed? The respective *.ts file is not always reported as "affectedFile" when I change the template file and that's why I was forcing it to re-run the diagnostics.

I can't use a cache to store the diagnostics since the diagnostics errors could be outdated.

Carlosamouco: I think I managed to find a simple way to make it work.

Carlosamouco: I think I managed to find a solution for this using the existing `FileCache`.

src/lib/ngc/compile-source-files.ts
```diff
@@ -159,11 +159,17 @@ export async function compileSourceFiles(
...
+    ) {
+      const angularDiagnostics =
+        (!affectedFiles.has(sourceFile) && sourceFileCache.get(sourceFile.fileName)?.angularDiagnostics) ||
+        angularCompiler.getDiagnosticsForFile(
+          sourceFile,
+          affectedFiles.size === 1 ? /** OptimizeFor.SingleFile **/ 0 : /** OptimizeFor.WholeProgram */ 1,
+        );
 
       allDiagnostics.push(...angularDiagnostics);
       sourceFileCache.updateAngularDiagnostics(sourceFile, angularDiagnostics);
```
alan-agius4, line 175: Consider the below, to improve readability

```suggestion
    if (affectedFiles.has(sourceFile)) {
      const angularDiagnostics = angularCompiler.getDiagnosticsForFile(
        sourceFile,
        affectedFiles.size === 1 ? /** OptimizeFor.SingleFile **/ 0 : /** OptimizeFor.WholeProgram */ 1,
      );

      allDiagnostics.push(...angularDiagnostics);
      sourceFileCache.updateAngularDiagnostics(sourceFile, angularDiagnostics);
    } else {
      const cachedDiagnostics = sourceFileCache.getAngularDiagnostics(sourceFile);
      if (cachedDiagnostics?.length) {
        allDiagnostics.push(...cachedDiagnostics);
      }
    }
```

Carlosamouco: We still need to account for cases where the file in `FileCache` was deleted by the file watcher due to a change or indirect change (in case of an update to a component template file) and diagnostics were lost/need to be re-evaluated.

```ts
const cachedDiagnostics = sourceFileCache.getAngularDiagnostics(sourceFile);
if (affectedFiles.has(sourceFile) || !cachedDiagnostics) {
  ...
}
```

As I said before, sometimes the component file is not reported as affected after a few saves to its template file and that's why I also check if it exists in `FileCache`, and if not, I know it has changed and deleted by the watcher and I need to re-run diagnostics.
Maybe the fact that it is not reported as affected it's a bug in the compiler but I am not sure...

Carlosamouco: Do you have any idea where the root cause of this problem is?

alan-agius4: Do you have a test case that I can test against?

Carlosamouco: I tested you PR but I am still able to reproduce the problem. I am going to try and create the test cases.

Carlosamouco: I have added a two tests that reproduce the issue.

The first one saves the same template file a few times, without making any change on it, expecting always an error but it ends up compiling successfully after a couple of saves.

In the second test that I added I tried to do multiple saves on a valid component file and then change it's input name expecting to get an error in a secondary entry point.

alan-agius4: thanks for this.

Carlosamouco: My first approach on this issue was to force a new version of the source file for unchanged files. While there might be an issue with the angular compiler, would it be possible to do this with a workaround? I really don't know where the problem is or a better way to solve it.

alan-agius4:
This is now obsolete due to recent changes.

PR #3000, opened 2025-03-12
feat: add DTS bundles for entry-points

### LINKED ISSUE #139: Flatten type definitions to a single *.d.ts file
## Type of Issue

- [ ] Bug Report
- [x] Feature Request

## Description

Currently, we output indivudual `*.d.ts` files for every source file

#### How To Reproduce

n/a

#### Expected Behaviour

In accordance with the flat ecmascript modules idea, we should create a flat `<library>.d.t.s` file that described the public API surface.

Inspiration taken from [ts-api-guardian](https://github.com/angular/ts-api-guardian/blob/master/lib/serializer.ts)

#### Version Information

```
ng-packagr: v1.2.0
```

alan-agius4:
This commit introduces DTS bundling.

Closes #139


integration/samples/api/specs/sourcemaps.ts
```diff
@@ -3,17 +3,18 @@ import * as fs from 'fs-extra';
...
-      sourceMap = fs.readJsonSync(path.resolve(__dirname, '../dist/src/angular.component.d.ts.map'));
+      sourceMap = fs.readJsonSync(path.resolve(__dirname, '../dist/types/sample-api.d.ts.map'));
     });
 
     it(`should exist`, () => {
       expect(sourceMap).to.be.ok;
     });
 
-    it('should point to the correct source path', () => {
+    // TODO: figure a way to make DTS maps work with bundling, or disable bundling when sourcemap is enabled.
```
devversion, line 16: I thought you got it working?

devversion (approved):

integration/samples/api/specs/sourcemaps.ts
alan-agius4: Yeah, but it does need an upstream change.

PR #3034, opened 2025-04-15
feat: expand browser support to widely available Baseline

dgp1130:
This changes the default `.browserslistrc` file to one generated from a Baseline "widely available" date using `bl2bl`, codifying the policy change from https://github.com/angular/angular/pull/60754 into `ng-packagr`.

This effectively pins the `.browserslistrc` file to a hard-coded configuration from a specific Baseline date, preventing it from automatically updating based on `caniuse-lite` versions managed in user workspaces, which should make Angular's tooling much more stable. Prior to this, expressions like "last 2 Chrome versions" were relative to the installed `caniuse-lite` version and could update at any time. By pinning `.browserslistrc`, we're ensuring that supported browsers only change during Angular major version updates, not arbitrary `caniuse-lite` updates.

The [Angular CLI implementation](https://github.com/angular/angular-cli/pull/30036) hard-codes the Baseline date and the uses `baseline-browser-mapping` to generate the `browserslist` configuration. We could copy that process here, however I didn't want to duplicate all the code and instead just hard-coded the `browserslist` configuration itself. As part of the release process, we can bump the date in Angular CLI, build the configuration, and then copy it over to `ng-packagr`. We already need to copy the config out for the docs anyways, so this doesn't significantly increase the toil involved.

Let me know if you'd rather I explore actually generating this information from a hard-coded date rather than inlining the `browserslist` config entirely. Also let me know if there's any particular ways I should be testing this.

Blocked on: https://github.com/angular/angular/pull/60754

### DIFF
--- a/src/lib/styles/stylesheet-processor.ts
+++ b/src/lib/styles/stylesheet-processor.ts
@@ -21,15 +21,10 @@ export class StylesheetProcessor extends ComponentStylesheetBundler {
     // By default, browserslist defaults are too inclusive
     // https://github.com/browserslist/browserslist/blob/83764ea81ffaa39111c204b02c371afa44a4ff07/index.js#L516-L522
     // We change the default query to browsers that Angular support.
-    // https://angular.io/guide/browser-support
-    (browserslist.defaults as string[]) = [
-      'last 2 Chrome versions',
-      'last 1 Firefox version',
-      'last 2 Edge major versions',
-      'last 2 Safari major versions',
-      'last 2 iOS major versions',
-      'Firefox ESR',
-    ];
+    // https://angular.dev/reference/versions#browser-support
+    (browserslist.defaults as string[]) = browserslist(undefined, {
+      path: require.resolve('../../../.browserslistrc'),
+    });
 
     const browserslistData = browserslist(undefined, { path: basePath });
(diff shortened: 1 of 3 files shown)


src/lib/styles/stylesheet-processor.ts
```diff
@@ -21,15 +21,10 @@ export class StylesheetProcessor extends ComponentStylesheetBundler {
     // By default, browserslist defaults are too inclusive
     // https://github.com/browserslist/browserslist/blob/83764ea81ffaa39111c204b02c371afa44a4ff07/index.js#L516-L522
     // We change the default query to browsers that Angular support.
-    // https://angular.io/guide/browser-support
```
alan-agius4, line 24: NIT: since the browserlist is static and not auto generated, might as well, put the supported browsers queries directly in `browserslist.defaults` instead of creating a new file.

alan-agius4 (approved):
LGTM

src/lib/styles/stylesheet-processor.ts
dgp1130: Had a chat offline and decided to keep the separate file just to make it easier to reference and discover when bumping browser versions during releases.

PR #3172, opened 2025-11-12
perf: reduce memory usage for multiple entry-points (non-watch)

### LINKED ISSUE #3168: Excessive memory usage during build with many secondary entry points
## Type of Issue

```
[x] Bug Report
[ ] Feature Request
```

## Description

When building libraries with a big amount of secondary entry points, they consume a large amount of memory when caching is enabled.

After upgrading our library from Angular 15 to 20, we noticed a huge memory increase. Our limit was set to 4GB, but even when increasing it to more than 24GB it still ran out of memory. It is a large internal component library with more than 270 entry points. Disabling the cache in angular.json reduced memory usage back to normal amounts and build passed faster.

#### How To Reproduce

Here is a minimal example with a freshly created project
https://github.com/plepa170/ng20-build-memory-issue

I created one new entry point with one component and some fluff data and duplicated it 400 times. The entry points are independent from each other.

Memory usage: 6GB with cache, 1.7GB without.

Tested the same example with some previous versions
Angular 15: 1.5GB
Angular 16: 12.4GB
Angular 18: 4.5GB

#### Expected Behaviour

I expect the memory usage to stay within reasonable amounts. 
In the specific given example, since the entry points are independent of each other, I would expect the amount of cache that needs to be held in memory to be low.

#### Version Information

```
ng-packagr: 20.3.0
@angular/*: 20.3.10
typescript: 5.9.3
rxjs: 7.8.2
node: 22.20.0
yarn: 1.22.5
```

alan-agius4:
reduce memory usage when building projects with multiple entry-points in non-watch mode.

Closes #3168

### DIFF
--- a/src/lib/ng-package/nodes.ts
+++ b/src/lib/ng-package/nodes.ts
@@ -4,7 +4,7 @@ import ts from 'typescript';
 import { FileCache } from '../file-system/file-cache';
 import { ComplexPredicate } from '../graph/build-graph';
 import { Node } from '../graph/node';
-import { by,  isInProgress, isPending } from '../graph/select';
+import { by, isInProgress, isPending } from '../graph/select';
 import { AngularDiagnosticsCache } from '../ngc/angular-diagnostics-cache';
 import { StylesheetProcessor } from '../styles/stylesheet-processor';
 import { DestinationFiles, NgEntryPoint } from './entry-point/entry-point';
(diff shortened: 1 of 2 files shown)


pumano:
@alan-agius4 looks like `dipose` is typo for `dispose`

alan-agius4:
Yes I’ll fix that

PR #3191, opened 2025-11-28
fix: set filename for declarations bundle on rollup

### LINKED ISSUE #3085: Library internal declaration merging fails since v20
We are using relative module declarations (`declare module "../xyz"`) in our projects to apply the open/closed principle to certain interfaces. For example, we might have a setup such as the following:

### options.ts

```typescript
export interface Options {
  globalOption: string;
}
```

### feat-a/options.ts

```typescript
export interface FeatAOptions {
  a: string;
}

declare module "../options" {
  interface Options {
    featA: FeatAOptions;
  }
}
```

And so on, for each feature. This leverages TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to allow us up to keep the `Options` interface closed to modification, but open to extension and to keep the feature specific configuration options declaration next to the rest of the feature specific code. This worked well up to Angular 19, although cross-library interface merging was a bit problematic, because you always had to reference the exact file into the library that contained the original interface declaration due to an [issue](https://github.com/microsoft/TypeScript/issues/12607) in TypeScript.

With Angular 20 cross library interface merging has become significantly easier, but intra-library mergin is now impossible. This is because Angular/ng-packagr 20 hoists all type declarations for the public API into the `index.d.ts` file instead of linking it up to a `public-api.d.ts` file and the individual files in their subdirectories in turn. As a result, other libraries can now reference the type for merging directly at the library level (which is great). 

On the other hand, though, the `declare module` blocks are copied into the `index.d.ts` file verbatim, which will obviously fail for relative paths, because those relative paths are no longer valid after hoisting.

You will find a [minimal reproduction](https://github.com/schoel-bis/ng-packagr-type-hoisting-issue) on GitHub. It has a library `lib-a` with two features (`feat_a` and `feat_b`), each of which merges its own options into the library level `ConfigOptions` interface. The `main` branch is the Angular 19 version, which works just fine (`npm i && ng build lib-a && ng serve`). The `ng20` branch uses Angular 20 and fails at the `ng serve` step with:

```
Application bundle generation failed. [0.925 seconds]

✘ [ERROR] NG9: Property 'featA' does not exist on type 'ConfigOptions'. [plugin angular-compiler]

    src/app/app.component.ts:12:22:
      12 │         <dd>{{ config.featA.featAOption1 }}</dd>
         ╵                       ~~~~~


✘ [ERROR] NG9: Property 'featB' does not exist on type 'ConfigOptions'. [plugin angular-compiler]

    src/app/app.component.ts:14:22:
      14 │         <dd>{{ config.featB.featBOption1 }}</dd>
         ╵                       ~~~~~
```

I am not sure what the correct fix is for this issue. Simply stripping relative `declare module` blocks and only keeping their contents could be an option. I am not sure, though, if that might break somebody else's code in ways that have not yet occurred to me. Otherwise, it might help to have an option to return to the previous behaviour on a per-library basis (e.g. via `ng-package.json`).

### LINKED ISSUE #3190: Relative declare module statements are wrongly rewritten since v21
Now that rollup-plugin-dts has released a new version that includes a fix for #3085 builds of libraries with relative `declare module` blocks work again with ng-packagr 20. With version 21, however, the resulting `.d.ts` files are invalid. The reason is that 7a5e8fbd51010db91d0a74bf6efb6c0eb6732a00 a) changes the name of the declaration file after it has been written by rollup and b) does not even provide a file name to rollup.

The latter means that rollup-plugin-dts will fall back to `index` as the default file name, which leads to the `declare module` blocks to be rewritten to `declare module "./index" {…}`. This worked fine in ng-packagr 20, because the declarations file was in fact called `index.d.ts` and thus the blocks referenced the correct file. After the mentioned commit, though, the blocks are still rewritten as `declare module "./index" {…}`, but the file has a different name and therefore the blocks reference another file (`index.d.ts`) in the same directory, which does not really exist.

This would be fixed if ng-packagr passed the target file name to rollup when bundling the `.d.ts` file.

schoel-bis:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

Sorry, no test yet. I wanted to quickly propose the approach and will add a test when I find the time.

## Description

After 7a5e8fbd51010db91d0a74bf6efb6c0eb6732a00 .d.ts bundles containing `declare module` statements with relative paths are no longer valid. This is because the change to rollup-plugin-dts that fixed ng-packagr/ng-packagr#3085 expects to find the target file name in options.file and otherwise falls back to index.d.ts.

This commit makes sure the full and proper file name is set in the options for rollup.

Fixes #3190

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/flatten/rollup.ts
+++ b/src/lib/flatten/rollup.ts
@@ -7,22 +7,31 @@ import { readCacheEntry, saveCacheEntry } from '../utils/cache';
 import * as log from '../utils/log';
 import { fileLoaderPlugin } from './file-loader-plugin';
 
-/**
- * Options used in `ng-packagr` for writing flat bundle files.
- *
- * These options are passed through to rollup.
- */
-export interface RollupOptions {
+interface RollupCommonOptions {
   moduleName: string;
   entry: string;
   entryName: string;
-  dir: string;
   cache?: RollupCache;
   cacheDirectory?: string | false;
   fileCache: OutputFileCache;
   cacheKey: string;
   sourcemap: boolean;
 }
+
+interface RollupOptionsDir extends RollupCommonOptions {
+  dir: string;
+}
+
+interface RollupOptionsFile extends RollupCommonOptions {
+  file: string;
+}
+
+/**
+ * Options used in `ng-packagr` for writing flat bundle files.
+ *
+ * These options are passed through to rollup.
+ */
+export type RollupOptions = RollupOptionsDir | RollupOptionsFile;
 
 let rollup: typeof import('rollup') | undefined;
 
@@ -76,7 +85,7 @@ export async function rollupBundleFile(
   const output = await bundle.write({
     name: opts.moduleName,
     format: 'es',
-    dir: opts.dir,
+    ...('dir' in opts ? { dir: opts.dir } : { file: opts.file }),
     inlineDynamicImports: false,
     hoistTransitiveImports: false,
     chunkFileNames: `${opts.entryName}-[name]-[hash]${outExtension}`,
--- a/src/lib/ng-package/entry-point/write-bundles.transform.ts
+++ b/src/lib/ng-package/entry-point/write-bundles.transform.ts
@@ -97,7 +97,7 @@ export const writeBundlesTransform = (options: NgPackagrOptions) =>
             entry: declarations,
             entryName: ngEntryPoint.flatModuleFile,
             moduleName: ngEntryPoint.moduleId,
-            dir: declarationsDir,
+            file: join(declarationsDir, `${ngEntryPoint.flatModuleFile}.d.ts`),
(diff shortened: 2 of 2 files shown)


alan-agius4 (changes requested):
This does not seem correct. We should always use `dir` as otherwise it would break cases where there are multiple dts chunks.

I suggest you provide a minimal reproduction/create a test.

schoel-bis:
My thinking was that since it was simply named index.d.ts before and placed in the root directory, we would probably only ever have a single d.ts chunk per entry point. I will provide a reproduction soon.

Gesendet von Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

PR #3266, opened 2026-03-29
fix: use correct this context in cacheCompilerHost method delegates

benpsnyder:
Part of https://github.com/ng-packagr/ng-packagr/pull/3252

## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

`cacheCompilerHost` wraps several `compilerHost` methods (`fileExists`, `getSourceFile`, `writeFile`, `readFile`, `readResource`) inside arrow functions, then delegates via `.call(this, ...)`. Because these are arrow functions, `this` resolves to the enclosing scope (likely `undefined` in strict mode) rather than the `compilerHost` instance. This works today only because the wrapped TS host methods don't currently dereference `this`, but it is incorrect and fragile.

This PR changes `.call(this, ...)` → `.call(compilerHost, ...)` so the delegated calls receive the proper context.

Additionally, this PR hardens the stylesheet-processing paths inside `readResource` and `transformResource`:

- Guards against `stylesheetProcessor` being `undefined` before calling its methods
- Uses null-coalescing for `referencedFiles` when spreading into arrays
- Replaces optional-chaining length checks with explicit truthiness checks for esbuild warnings/errors
- Adds a `?? ''` fallback on `cache.content` return and an explicit `: Promise<string>` return type on `readResource`

### Why

Arrow functions do not bind their own `this` — they inherit it from the enclosing lexical scope. When the returned host object's methods are called, `this` inside them is **not** the host object but whatever `this` was when `cacheCompilerHost` was invoked (typically `undefined` in strict mode). Passing that to `.call()` means the delegated method receives the wrong context. If a future TypeScript compiler-host implementation starts referencing `this` internally, these calls would break silently.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/lib/ts/cache-compiler-host.ts
+++ b/src/lib/ts/cache-compiler-host.ts
@@ -54,7 +54,7 @@ export function cacheCompilerHost(
     fileExists: (fileName: string) => {
       const cache = sourcesFileCache.getOrCreate(fileName);
       if (cache.exists === undefined) {
-        cache.exists = compilerHost.fileExists.call(this, fileName);
+        cache.exists = compilerHost.fileExists.call(compilerHost, fileName);
       }
 
       return cache.exists;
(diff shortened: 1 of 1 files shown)


benpsnyder:
@alan-agius4 ready

alan-agius4 (dismissed):
LGTM, thanks

alan-agius4 (approved):

PR #3288, opened 2026-04-18
Fix: handle package.json files with export subpaths

### LINKED ISSUE #3287: Failure writing package.json mainfest when exports subpath is a string
If the package.json file defines export subpaths as strings, that get changed by ng-packagr, an error is thrown.

```
✖ Writing package manifest
Cannot create property 'default' on string './package.json'

```

### package.json
```
{
  "name": "@sample/package-exports",
  "description": "A sample library using package.json subpath exports",
  "version": "1.0.0-pre.0",
  "private": true,
  "repository": "https://github.com/ng-packagr/ng-packagr.git",
  "peerDependencies": {
    "@angular/core": "^4.1.2",
    "@angular/common": "^4.1.2",
    "@angular/common": "^4.1.3",
    "@angular/core": "^4.1.3",
    "rxjs": "^5.0.1"
  },
  "exports": {
    "./package.json": "./package.json"
  }
}
```

This is because the write-package.transform reads the exports and believes they are objects but if it is a string and a property is written to a string then an error is thrown in strict mode.

https://github.com/ng-packagr/ng-packagr/blob/c74a6429a887fc2629b18ed3bebf86a1ec968026/src/lib/ng-package/entry-point/write-package.transform.ts#L334-L353

soda0289:
Fixes #3287

## I'm submitting a...

```
[X] Bug Fix
[ ] Feature
```

## Checklist

## Description

When a package.json file defines export subpaths, that will be overwritten by ng-packagr, it should be able to handle instances when the export is a subpath string or a conditional export object.

The fix will check the type of the subpath export and if it is a string display a warning to a user and set the export to empty object. If it is an object it will delete any existing subpath exports that are being overwritten to ensrue the order of exports is maintained. The `default` export must always come last even when it is already defined in package.json

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -313,7 +313,7 @@ function checkNonPeerDependencies(
   }
 }
 
-type PackageExports = Record<string, ConditionalExport>;
+type PackageExports = Record<string, ConditionalExport | string>;
 
 /**
  * Type describing the conditional exports descriptor for an entry-point.
@@ -333,6 +333,10 @@ function generatePackageExports(
 
   const insertMappingOrError = (subpath: string, mapping: ConditionalExport) => {
     exports[subpath] ??= {};
+
+    if (typeof exports[subpath] === 'string') {
+      exports[subpath] = { default: exports[subpath] };
+    }
     const subpathExport = exports[subpath];
 
     // Go through all conditions that should be inserted. If the condition is already
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -9,6 +9,7 @@ import { transformFromPromise } from '../../graph/transform';
 import { colors } from '../../utils/color';
 import { copyFile, mkdir, rmdir, stat, writeFile } from '../../utils/fs';
 import * as log from '../../utils/log';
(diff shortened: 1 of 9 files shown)


src/lib/ng-package/entry-point/write-package.transform.ts
```diff
@@ -333,6 +333,10 @@ function generatePackageExports(
 
   const insertMappingOrError = (subpath: string, mapping: ConditionalExport) => {
     exports[subpath] ??= {};
+
+    if (typeof exports[subpath] === 'string') {
+      exports[subpath] = { default: exports[subpath] };
```
soda0289: This will break the order of conditional exports and cause default to be first ignoring other mappings.

soda0289:
Pushed change to PR that now just prints a warning when the subpath export is a string since it will always overwrite default. Also deletes existing subpath export conditions when they are being overwritten to ensure the order is maintained and `default` is always last.

alan-agius4 (changes requested):
Thanks for this. 

Instead of creating an integration test, for this case a unit test is enough. (https://github.com/soda0289/ng-packagr/blob/f20a3009b43cf231477992c44fbc75cdd303bef5/src/lib/ng-package/entry-point/entry-point.spec.ts)

soda0289:
Moved the generatePackageExports() method out of the transform to make unit test easier to write.

soda0289:
Pushed changes to fix lint and format issues

alan-agius4 (approved):

PR #3320, opened 2026-06-02
fix: invalidate angularDiagnosticCache for html changes

mver-al:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

In watch mode with an angular component consisting of a TS and template HTML file: after introducing an issue in the template by changing the TS code and then changing the HTML to resolve the issue, the issue would not disappear and watch mode would need to be restarted manually. After some investigation I found this was due to the `angularDiagnosticCache` only being invalidated on TS updates (`affectedFiles`). In this PR the cached diagnostics for changed HTML files will be deleted as they need to be regenerated.

I have made the fix for 21.x as that is where I am affected, but it should also be applicable for the main branch.

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/watch/basic.spec.ts
+++ b/integration/watch/basic.spec.ts
@@ -33,7 +33,7 @@ describe('basic', () => {
         });
       });
 
-      it('should recover from errors', done => {
+      it('should recover from template errors', done => {
         harness.copyTestCase('invalid-type');
 
         harness.onComplete(() => {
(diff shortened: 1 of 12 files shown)


alan-agius4:
Thanks for this change. Kindly add a test in Please add a test in `integration/watch/basic` that validates the change or provide a minimal reproduction of issue.

mver-al:
Hi Alan, I ran into the issue when converting back and forth between `@Input()` <-> `input()`

I have made a minimal example [here ](https://github.com/mver-al/ng-packagr-watch-issue)(also with Angular v. 21 as that is what I am using). To reproduce, with watch mode running:
first convert in my-lib.ts:
```
x = input(5)
```
to 
```
@Input()
x = 5;
```

Then compilation will fail on the template.

now change in the template
```
<p>my-lib works! {{x()}}</p>
```
to
```
<p>my-lib works! {{x}}</p>
```
resolving the error, and observe that the errors persist in the watch mode output.

alan-agius4 (approved):
LGTM thanks

alan-agius4:

- main: e7d8e38a30c6197a6f2f450d41b1752cce9360e2
- 21.2.x: 231243368b6008d5be25f9c881c99a9c893ae533

PR #3329, opened 2026-06-08
fix(ng-packagr): prevent concurrent compilation and stale cache results

alan-agius4:
This PR ports the changes from angular-cli #33318 to `ng-packagr` to resolve similar concurrent compilation and race condition issues during stylesheet processing.

### Changes:
- **Cache**:
  - Store and share in-flight creator promises (`#requests`) to prevent multiple concurrent creations for the same key.
  - Track writes during async gaps (`#writeCounts`) to abort and restart `getOrCreate` if a write occurred during the await gap.
  - Track active gets (`#pendingGets`) to clean up write count tracking and avoid memory leaks.
- **BundlerContext**:
  - Share the active bundle promise (`#activeBundlePromise`) to prevent concurrent executions of the esbuild build process.
  - Safely track and check the `#disposed` state when context is created asynchronously, to abort if it was disposed.
- **Tests**:
  - Added unit tests for the updated `MemoryCache` implementation.

### DIFF
--- a/src/lib/styles/bundler-context.ts
+++ b/src/lib/styles/bundler-context.ts
@@ -61,6 +61,8 @@ export class BundlerContext {
   #esbuildContext?: BuildContext<{ metafile: true; write: false }>;
   #esbuildOptions?: BuildOptions & { metafile: true; write: false };
   #esbuildResult?: BundleContextResult;
+  #activeBundlePromise?: Promise<BundleContextResult>;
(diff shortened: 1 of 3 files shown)


alan-agius4:

- main: e5d77a7837fab18f7a200126df7f8fd703f55462
- 22.0.x: c4fd75f93ef18fb31c88c0bcb21e971add052eb1

PR #3344, opened 2026-06-29
fix(ng-packagr): generate specific .npmignore entries for secondary e…

### LINKED ISSUE #3343: Nested schematics and builders package.json's should not be ignored
I'm authoring an Angular package that also contains **schematics** and **builders**.

The final main **package.json** generated by **ng-packagr** now has a `type: module` which is desired behavior.

This forces schematics and builders to have their own **package.json** with `type: commonjs` (which is fine too) to avoid errors like this one when running them:
```
An unhandled exception occurred: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '@hug/ngx-g11n/package.json' contains "type": "module". 
To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

### The issue

This is completely fine expect that **ng-packagr** is also generating an **.npmignore** file in the dist folder which ignores entirely the nested package.json's from schematics and builders:

```
# Nested package.json's are only needed for development.
**/package.json
```

### Temporary fix

Modify the `.npmignore`:

```
# Nested package.json's are only needed for development.
**/package.json
!schematics/package.json
!builders/package.json

```

alan-agius4:
…ntry-point package.json files

Previously, a wildcard `**/package.json` entry was added to the `.npmignore` file of the primary entry-point to prevent publishing nested `package.json` files. However, this wildcard is too broad and can inadvertently exclude legitimate `package.json` files that should be packaged (such as those in assets or fixtures). This commit resolves the issue by targeting only the specific, auto-generated `package.json` files of secondary entry-points. We now compute their paths relative to the primary destination and list them individually in the `.npmignore` file, ensuring other nested `package.json` files are not affected.

Closes #3343

### DIFF
--- a/src/lib/ng-package/entry-point/write-package.transform.ts
+++ b/src/lib/ng-package/entry-point/write-package.transform.ts
@@ -48,10 +48,26 @@ export const writePackageTransform = (options: NgPackagrOptions) =>
         spinner.start('Writing package manifest');
         if (!options.watch) {
           const primary = ngPackageNode.data.primary;
-          await writeFile(
-            path.join(primary.destinationPath, '.npmignore'),
-            `# Nested package.json's are only needed for development.\n**/package.json`,
-          );
+
+          const nestedPackageJsons: string[] = [];
+          // Secondary entry-points package.json
+          for (const { data } of entryPoints) {
+            if (!data.entryPoint.isSecondaryEntryPoint) {
+              continue;
+            }
+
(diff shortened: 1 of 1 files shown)


alan-agius4:

- main: a63fbc0dc053d1c69059067054038e312fe1d029
- 22.0.x: 44bc3a7a5cf07072b2c5556a6e2af6aa22d6e600

PR #3352, opened 2026-07-20
fix(ng-packagr): log correct diagnostics array for warnings

### LINKED ISSUE #3346: Non-error diagnostics (warnings) are never printed: wrong array logged in compileSourceFiles
## Summary

`compileSourceFiles()` in `src/lib/ngc/compile-source-files.ts` silently swallows every non-error compiler diagnostic (warnings, suggestions, messages, including Angular's template `extendedDiagnostics` like NG8107 "optional chain not nullable" or NG8102 "nullish coalescing not nullable"). Any build that only has warnings, no errors, finishes with zero console output about them.

## Root cause

```ts
const otherDiagnostics = [];
const errorDiagnostics = [];
for (const diagnostic of allDiagnostics) {
  if (diagnostic.category === ts.DiagnosticCategory.Error) {
    errorDiagnostics.push(diagnostic);
  } else {
    otherDiagnostics.push(diagnostic);
  }
}

if (otherDiagnostics.length) {
  log.msg(formatDiagnostics(errorDiagnostics)); // <-- wrong variable
}

if (errorDiagnostics.length) {
  throw new Error(formatDiagnostics(errorDiagnostics));
}
```

The `if (otherDiagnostics.length)` check correctly detects that there are non-error diagnostics to show, but the line inside logs `errorDiagnostics` instead of `otherDiagnostics`. When a build only has warnings, `errorDiagnostics` is empty, so `formatDiagnostics([])` returns nothing and `log.msg('')` prints nothing. The warnings just get thrown away right after being computed.

This line goes back to 0daa33e (Sept 27 2021), and that commit's own message is "fix: don't exit with non zero error code on non error compiler diagnostics". So the goal was to keep showing warnings without failing the build on them. Only the "don't fail" half made it in. Still there on main right now.

## How to reproduce

Tested on ng-packagr 17.3.0 (same code on main), Angular 17.3, workspace with `strictTemplates: true`.

A component with an input typed through a definite assignment assertion:

```ts
@Input() public contact!: ContactModel;
```

and its own template doing an optional chain on that same input:

```html
@if (contact?.objetType) { ... }
```

`strictTemplates` + `extendedDiagnostics` correctly flags this as NG8107 (category Warning), since the declared type of `contact` doesn't include `undefined` once you assert it with `!`.

Build the lib with the Angular CLI's ng-packagr builder (`ng build`, tried both `compilationMode: partial` and `full`, no difference):

```
Compiling with Angular sources in Ivy partial compilation mode.
✔ Compiling with Angular sources in Ivy partial compilation mode.
✔ Built ...
```

Nothing about the warning shows up anywhere, build exits 0.

Now compile that same source directly through the Angular CLI application builder instead (an app that resolves the lib to source instead of `dist/`, or just `ng serve`/`ng build` on an app containing the component) and the same NG8107 shows up fine, with file and line:

```
▲ [WARNING] NG8107: The left side of this optional chain operation does not include 'null'
or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator.

    badge.component.html:1:14:
      1 │ @if (contact?.objetType) {
        ╵               ~~~~~~~~~
```

Same diagnostic, same `ts.Diagnostic`, computed the same way by ngtsc both times. The only difference is which pipeline prints it.

We ran into this in a monorepo with 200+ Angular libraries built through ng-packagr. The `@Input() foo!: Type` + `?.`/`??` pattern shows up in 400+ places across 200+ files there, so it's not an edge case on our end. Every one of those warnings has been silently dropped on every library build for years, and we only noticed once we started compiling library sources directly through the app builder for a faster local dev loop.

## Fix

```diff
 if (otherDiagnostics.length) {
-  log.msg(formatDiagnostics(errorDiagnostics));
+  log.msg(formatDiagnostics(otherDiagnostics));
 }
```

alan-agius4:
When non-critical compiler warnings/diagnostics are collected, the code was mistakenly formatting and logging the `errorDiagnostics` array instead of the `otherDiagnostics` array. Because `errorDiagnostics` is either empty or contains errors that would be thrown anyway, warnings were never actually printed to the console.

This change corrects the variable to format and log `otherDiagnostics`.

Fixes #3346

### DIFF
--- a/src/lib/ngc/compile-source-files.ts
+++ b/src/lib/ngc/compile-source-files.ts
@@ -182,7 +182,7 @@ export async function compileSourceFiles(
   }
 
   if (otherDiagnostics.length) {
-    log.msg(formatDiagnostics(errorDiagnostics));
+    log.msg(formatDiagnostics(otherDiagnostics));
   }
 
   if (errorDiagnostics.length) {


alan-agius4:

- main: 401d5e71f2f9a8c7131415956f22306423363e49
- 22.0.x: 9d1c0cd01be79f25765751da1b768f4848ecc199

PR #3367, opened 2026-08-02
build: replace ts-node, jasmine-ts with tsx

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

Replace ts-node and jasmin-ts with tsx

I've been told thats state of the art in 2026... :)

```
[ ] Yes
[x] No
```


alan-agius4 (changes requested):
Node.js now supports running TypeScript files directly. It would be great if we could use that built-in capability instead of relying on `tsx`.

Is it possible or there are some additional blockers?

dherges:
> Node.js now supports running TypeScript files directly. It would be great if we could use that built-in capability instead of relying on `tsx`.
> 
> Is it possible or there are some additional blockers?

I read that Node.js runs TypeScript files with a _Type Stripping_ approach. Because it only removes types and does not transform code, there must not be `enum` or TypeScript decorators in the sources (and some other TS-only features are not supported).

I guess we would need to split this:
1. compile ng-packagr from ts to js
2. build the `integration/samples` through the compiled ng-packagr
3. run the integration/**/spec/*.ts` files (for integration only) through node.js natively

And find an alternative solution for npm tasks like:
- ‎"integration:samples:dev"‎
- "test:specs"

Reasoning: tasks referencing ng-packagr ts sources (not the compiled js) will likely break. Migrating the ng-packagr source code to strip any non-supported TypeScript feature such as `enum` would be an option.

Do you have experiences with this in the Angular core or Angular CLI? Where does it sit on the team's roadmap?

alan-agius4 (approved):
Let's stick with tsx for now, as introducing this will require a refactor to handle file extensions. That said, we generally don't use tsx in Angular because our compilation and testing rely on Bazel

alan-agius4:

- main: b592bcb74066d91f8638a20ee9af5f2dcf21768c

PR #3380, opened 2026-08-07
feat(ng-packagr): transform supported browsers to targets using lowest version

alan-agius4:
Transform browserslist results to only retain the lowest version for each browser target to avoid issues with esbuild and rolldown when multiple versions of the same engine are specified.

Also expand `ESBUILD_SUPPORTED_BROWSERS` to include all supported browsers as `ReadonlySet<string>` and ensure lone major versions are properly formatted.


alan-agius4:

- main: 148f6408716e363ab9e7cc84db582dae256a8033

PR #3418, opened 2026-09-10
test: prefer vitest over jasmine in 2026

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

Run the `.spec.ts` files through vitest.

Replaces jasmine. Reduces dependency on `tsx` as side effect.

Removes stranded `mocha` references.

```
[ ] Yes
[x] No
```


dherges:
Missed the vscode luanch configs

package.json
```diff
@@ -112,13 +111,13 @@
...
-    "integration:samples": "tsx integration/samples.ts",
-    "integration:samples:dev": "tsx --tsconfig src/tsconfig.packagr.json ./integration/samples.dev.ts",
-    "integration:specs": "tsx ./node_modules/jasmine/bin/jasmine.js --config=integration/tsconfig.specs.json \"integration/samples/*/specs/**/*.ts\"",
     "integration:consumers": "integration/consumers.sh",
     "integration:consumers:ngc": "ngc --project integration/consumers/tsc/tsconfig.json",
-    "integration:watch:specs": "tsx ./node_modules/jasmine/bin/jasmine.js --config=integration/tsconfig.specs.json --random=false \"integration/watch/*.spec.ts\"",
-    "test:specs": "tsx ./node_modules/jasmine/bin/jasmine.js --config=src/tsconfig.specs.json \"src/**/*.spec.ts\"",
+    "integration:samples": "tsx integration/samples.ts",
+    "integration:samples:dev": "tsx --tsconfig src/tsconfig.packagr.json ./integration/samples.dev.ts",
+    "integration:specs": "vitest run --config vitest.integration.config.ts",
```
alan-agius4: Looks like this is also running the watch spec files. In fact the number of specs executed is not the same.

vitest.config.mts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vitest/config';
```
alan-agius4, line 1: Use an `.mts` extension to avoid Vite 5 deprecation warnings.

vitest.integration.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    include: ['integration/samples/*/specs/**/*.ts', 'integration/watch/*.spec.ts'],
+    globals: true,
+    sequence: {
+      shuffle: false,
```
alan-agius4: This should only be disabled for the watch specs.

vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
```
alan-agius4: Use an `.mts` extension to avoid Vite 5 deprecation warnings.

alan-agius4: Disable test isolation, which should speed up the tests as we do not rely on side effects.

vitest.config.mts
alan-agius4, line 1: Disable test isolation, which should speed up the tests as we do not rely on side effects.

alan-agius4, line 1: Can we avoid multiple configuration files and just pass the include as an arg?

dherges: Reduced to just one vitest config file

package.json
dherges: Updated and verified the number of test files being run by vitest matches the actual count of `*.spec.ts` and `spec/*.ts` files in the folders

vitest.config.ts
dherges: Speeds it up by 40%+ to 80%+

Did some rough estimations by local execution in a GitHub codespace, passing `--isolate` vs. `--no-isolate` flag to vitest

vitest.config.mts
dherges: Good catch, thanks!

package.json
dherges: Moved the 3 glob patterns to the `include` in main `vitest.config.mts`.

Added path-based filter for the test suite to the vitest arguments in main `package.json`.

alan-agius4 (approved):

PR #3425, opened 2026-09-10
test: remove dependencies on chai, fs-extra

dherges:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

Reducing code base drift between ng-packagr and the angular-cli codebase.

Removed `chai` and `fs-extra` from `devDependencies` as they are no dependencies in the angular-cli repo.

Removing usages of `tsx` moves the ng-packagr codebase closer to the angular-cli repo.

However, while midway through it, I realized that `vitest` is not any dependency in the angular-cli, either. Oops. This PR might turn out to be a step in the wrong direction. To the good points: `tsx` is now just used for 2 scripts to build the samples libs:
1. `integration:samples` can be replaced with plain js execution, now.
2. `integration:samples:dev` is the last one blocking the removal of `tsx`.

The 2nd one runs ng-packagr from `*.ts` sources building the samples libs. One option to replace it is `vite-node` (which isn't used by the angular-cli either). For another option, I would need to give it one more thought iteration. Building the samples libs in plain js while having `tsc --watch` in the background is an option.

Would love to hear your thoughts.

```
[ ] Yes
[x] No
```


vitest.config.mts
```diff
@@ -4,6 +4,5 @@ export default defineConfig({
   test: {
     include: ['integration/samples/*/specs/**/*.ts', 'integration/watch/*.spec.ts', 'src/**/*.spec.ts'],
     isolate: false,
-    globals: true,
```
dherges, line 7: Removed `globals: true` and migrated test files to explicit `import { .. } from 'vitest';`. However,`vitest` is not a dependency in the angular-cli repo. I think it's still a good to move by moving the ng-packagr closer to current state of art in the modern Angular.

Holding on for a moment: do we want to go this direction?

alan-agius4: The Angular repository uses Jasmine for testing, which shouldn't impact this project. Since globals are already disabled, we should also remove `vitest/globals` from `types` in [`tsconfig.json](https://github.com/ng-packagr/ng-packagr/blob/0dc2d5d130c1ed43436b6f197ca774882f7cb3d7/tsconfig.json#L22)

alan-agius4 (approved):

alan-agius4 (approved):

alan-agius4:
@dherges can you please rebase and fix the conflict?

dherges:
> @dherges can you please rebase and fix the conflict?

done.

alan-agius4 (approved):

PR #3435, opened 2026-09-17
fix(ng-packagr): only reject a parallel index.ts when it is imported

thekhegay:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Since #3388 (22.2.0-rc.0) the build fails for any entry point that has an `index.ts` next to a non-index `entryFile`, even when nothing imports it, e.g. an `index.ts` that only does `export * from './public-api'`. 22.1.1 builds these fine.

Before #3388 the check looked for `index.ts` in the program, so it only failed when the file was actually imported. This moves the check after the import scan and uses the visited files, which restores that.

```
[ ] Yes
[x] No
```


alan-agius4 (approved):
LGTM

thekhegay:
Thanks. The commit message check is red because the body is under the 20 character minimum; the commit only has a subject. I can add a body, but force-pushing drops the approved workflow run and it would need approving again, so tell me which you prefer.

thekhegay:
For context on the timing: I keep my library on the current Angular, and 22.2 is already in rc. With this regression every entry point that has an `index.ts` next to a non-index `entryFile` fails to build, so ngwr (227 secondary entry points, all of that shape) can't move to 22.2 at all once it is out.

PR #3438, opened 2026-09-18
fix(ng-packagr): scope Sass package resolution caching for stylesheets in node_modules

thekhegay:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

Since dbea592a7 a bare Sass specifier is cached as `url` alone, so the resolution made for one stylesheet is handed to every other stylesheet in the build. Depending on the order the stylesheets compile in, a stylesheet is compiled against a copy of the package it cannot see, or a stylesheet that should fail resolves someone else's copy, or a cached `null` makes a stylesheet that can resolve the package fail with `Can't find stylesheet to import.` An ordinary transitive version conflict is enough. 22.1.1 is correct on the same fixture, 22.2.0-rc.0 is not.

Package resolutions and package roots are now qualified with a scope, following the design agreed on angular/angular-cli#34119: a stylesheet inside `node_modules` uses the root of its enclosing package, so every file of a package shares one resolution while a dependency's nested copy stays isolated, and every other stylesheet uses the working directory, so a library's component stylesheets still share one resolution. The scope comes from the stylesheet path alone, with no file system access, and a non-`file:` containing URL resolves from the working directory instead of throwing.

The perf win of that commit is kept: over 900 stylesheets in 900 directories, 3 resolver calls, the same as the rc as shipped, against 2700 for the containing-URL key that preceded it. On a dependency with 40 subfolders importing the same package it is also 3, against 120 when the scope was the stylesheet's own directory.

Each spec fails when the part of the change it covers is reverted. `pnpm lint`, `pnpm test:specs`, `pnpm build`, `pnpm integration:samples` and `pnpm integration:specs` all pass.

The same change for `@angular/build`, which carries this file nearly verbatim, is in angular/angular-cli#34119.

```
[ ] Yes
[x] No
```


alan-agius4 (approved):
LGTM

